PR #8, opened 2015-06-22
Use AlphaSight's engineer page instead of the main dot com

jjbohn:
It's a better description of the engineer team and our tools intead of the main company website. Makes more sense to me to send people here.


doomspork:
Sounds good, thanks for the change :+1:

PR #12, opened 2015-07-22
Pinterest uses Elixir in production.

jparise:


doomspork:
Wow, this is awesome news @jparise!  Do you mind me asking what you use it for?  I'm sure many people would love to hear more.

jparise:
A small but growing number of our internal services are written in Elixir.  We have an [ETS](http://www.erlang.org/doc/man/ets.html)-based distributed rate limiting service, for example.

doomspork:
Thanks for the response @jparise, it's great to hear how Pinterest is using it.

On a separate note: I'm a co-organizer of the [Bay Area Elixir Meetup](www.meetup.com/Bay-Area-Elixir-Meetup) and we'd love to have you and your co-workers join us for our next meetup if you weren't planning on it already.

Thanks again.

PR #39, opened 2016-01-16
Add Red Panthers

manusajith:
@doomspork Thanks for curating the list of companies working on Elixir.


doomspork:
Thank you @manusajith :+1:

PR #41, opened 2016-01-21
Add Mirego

simonprev:
Thanks for making this list!

We use Elixir in many internal tools at Mirego. We are also actively contributing to open-source projects on GitHub.

We also built our open-source showcase page using Elixir and Phoenix (https://github.com/mirego/mirego-open-web)


doomspork:
Awesome!  Thank you @simonprev :+1:

PR #52, opened 2016-02-24
Add CrankWheel to the list :)

joisig:


doomspork:
:boom: 

Thanks @joisig :grinning:

PR #57, opened 2016-03-02
Adding Wombat Security

veverkap:
We are using Elixir in production for heavy data lifting


doomspork:
That's really cool @veverkap, thanks!

PR #58, opened 2016-03-04
Add Stampery

aesedepece:
We are Stampery strongly rely on Elixir and Erlang/OTP for our patent-pending "BTA" solution for scalable and distributed bitcoin blockchain data recording.


doomspork:
That's really cool @aesedepece, I hope you all write more about your usage! :+1:

PR #61, opened 2016-03-14
Remove Fyber.

carlosmoutinho:
While we are happy to see all the interest and fantastic community around Elixir we are not currently using it in a productive environment; and the position that mentioned Elixir is no longer open. In order to not mislead anyone we ask that the entry for Fyber is removed.


doomspork:
Sounds good @carlosmoutinho :+1:

PR #66, opened 2016-04-19
Added link to Madrid Elixir meetup

javiercr:
We've started co-organizing the first Elixir meetup in Madrid (the biggest one in Spain so far).


doomspork:
Awesome work @javiercr 👍

PR #70, opened 2016-04-27
HTTPS for Codenian doesn't work.

pragmaticivan:
Adding `http` instead.


doomspork:
Thank you @pragmaticivan 👍

PR #89, opened 2016-07-26
remove BT, no longer using Elixir

bordeltabernacle:


doomspork:
Thanks for the update @bordeltabernacle.  Would you care to share (for my own edification) why you moved away from Elixir?

bordeltabernacle:
Sure @doomspork 
Initially I had intended to use Phoenix for an internal web application, and really took to functional programming, after getting bogged down by the *magic* of OO & Django.  Subsequently I also started to use Elixir in my network programmability & automation projects.  
The internal web app project has stalled, and the networking industry has really settled on Python, and to a lesser extent, Go, and it didn't make sense to continue using an immature language like Elixir.  Also Python, and imperative programming, is just easier to expose to network engineers who do not want to get too deep into programming.
Elixir is a fantastic language, just now it's the wrong tool for the job.

doomspork:
Thank you @bordeltabernacle 😀

PR #90, opened 2016-07-27
Add Code Corps

JoshSmith:
@doomspork thanks for putting this list together! Saw it mentioned in the [Operation Code](https://operationcode.org/) Slack room.

I've added everything on the list except for how Elixir is used. Seems to be an open issue about whether that would make the list too verbose. We're in the process right now of migrating our entire API to Phoenix, but I do think that it could be a bit verbose expanding on the use of Elixir.

Let me know if you want that changed anyway and I'd be happy to make it.

Also happy to squash these commits if need be.

Thanks!

- [x] Company name
- [x] Company description
- [x] GitHub link
- [ ] How Elixir is used
- [x] Sorted alphabetically


doomspork:
You're welcome @JoshSmith, thanks for adding Code Corps!   RE: Elixir Usage, yeah, that's still on the fence.

Thank you again 👍

PR #99, opened 2016-08-19
Add FRED

alfredbaudisch:
Add [FRED](http://fredapp.com.br) to the list of companies.


doomspork:
Sounds awesome @alfredbaudisch 👍

alfredbaudisch:
Thanks @doomspork ! We are growing really fast and for that reason we are hiring Brazilian Elixir devs!

PR #224, opened 2017-10-30
Add Stuart Delivery

schurig:
Added Stuart


doomspork:
Thanks for the addition @schurig 👍

PR #239, opened 2017-11-26
Use TOML format instead Markdown

sprql:
I'd like to propose using TOML format for the companies list.
This format easy to maintain and to read, and easy to use for different kinds of automation.

```
[[company]]
name = "Example"
section = "Gaming"
url = "http://www.example.com/"
github = "https://github.com/example"
description = "Example is a company"
```

It's easy to generate any formats from this source.
It's also possible to have list of sections, and produce different lists.

For example I created [self contained HTML page](https://github.com/sprql/elixir-companies/blob/gh-pages/index.html) which uses this list and rendered it as a table.
And then I run it using github pages: [sprql.github.io/elixir-companies/](https://sprql.github.io/elixir-companies/)

It's possible to build README.md using CI as well.

What do you think?


doomspork:
@sprql this is a wonderful idea!  I'm all on board so long as we can provide an easy way for people to browse the data (much like the site you generated which is aligned with some plans I already had).

stevedomin:
Love the idea as well!

I think we need to do something about the section/industry otherwise I think it's going to become messy very quickly.

@doomspork anything I can do to help?

stevedomin:
@sprql: I added 3 companies: #240, #241, #242

doomspork:
@sprql this is wonderful!  Thank you again for taking the initiative on this.

@stevedomin I'm open to any ideas you've got.  I pinged @burden, a buddy of mine last night, about potentially throwing some CSS at this and make it easier to navigate.  That got me thinking about the possibility of flagging businesses that are hiring, especially those for whom a job link can be found.

This is really exciting and something I'd wanted to do for awhile with this project!

I was playing with some ideas for a logo, what do you think of something simple like this? 
image

elixir-companies.toml
```diff
@@ -0,0 +1,1362 @@
+[[company]]
```
stevedomin: Should we add the list of sections at the top?

We could use this for a test that runs in CI and check all the companies against these pre-defined sections.

stevedomin: It will also make it easier for people to know which section to use for the company they're adding.

stevedomin:
@sprql two more companies: #243 #244 

@doomspork should we aim to merge this asap to save some work for @sprql?

sprql:
@stevedomin I updated toml file, don't worry about work, I use [parser](https://github.com/sprql/elixir-companies/blob/feature/parser/readme_parser.ex) to build this file.

I changed field name from `section` to `industry` and added the list of all industries.

And add another variant of [layout](https://sprql.github.io/elixir-companies/table.html).

What are you think?
I can help if needed.

doomspork:
Sorry for the delay @sprql, just getting home from a long business trip.  I think your latest layout is a great start.  We can always make improvements from there.  I did merge in a few new businesses, could you add them to the TOML file and then I'll get this merged.

elixir-companies.toml
doomspork: I was _just_ thinking about this @stevedomin (sorry for the delay, was away on travel).  Perhaps we could rely on the PR template to guide them in submitting a company?

doomspork:
Thank you again @sprql!!

stevedomin:
This is super cool!

PR #255, opened 2017-12-20
Update Nebo #15 description

AndrewDryga:
We have changed our model and went distributed :).


doomspork:
Thanks for the update @AndrewDryga 😁

PR #279, opened 2018-03-30
Fix city, state for CargoTel

starbelly:
Yeah... I did that 😦  In my defense we are distributed and I never think about the location of HQ. It's also on a street called Boston Ave. I have excuses for days 😄


doomspork:
No worries @starbelly 😁

PR #290, opened 2018-05-16
fix links

dnsbty:
The Jekyll template links to the company websites and GitHub accounts aren't working. The company websites appeared to just be a typo.

The GitHub links are broken because the template appears to be expecting a GitHub username. I used the following regex pattern in the company list to verify that all of the :github fields are currently URLs.
```regex
github: ((?!http))
```
It might make sense to convert them all over and display the usernames instead of just `GitHub`. For now though, this should fix the links and that could be done in a separate PR.


dnsbty:
Looks like I was too slow with my PR and you beat me to it with #289

doomspork:
Dang!  Sorry @dnsbty 😞 

Thank you though!

PR #306, opened 2018-05-16
Fix the URL to the list

CraigCottingham:


doomspork:
Thank you @CraigCottingham, good catch 👍

PR #307, opened 2018-05-17
Update Alembic entry

joshprice:
- Use https
- Add location
- Tweak description


doomspork:
Thank you @joshprice 😁

PR #316, opened 2018-06-03
add chaince to elixir-companies

seamon:
add chaince to elixir-companies, including url, location, description, etc.


doomspork:
Thank you for the addition @seamon!

PR #321, opened 2018-06-11
remove companies with failing urls

jhchen:
Diacode and Govannon SSL certs are expired and Entelios's website does not respond at all. For Diacode the url could also just be changed to Github or just have them add themselves again if/when their SSL cert gets renewed.


doomspork:
Thank you @jhchen ❤️

PR #331, opened 2018-07-09
Update companies.yml

davneet4u:
Added teachoo to the companies list


doomspork:
@davneet4u do you want to add a link to the job posting?

davneet4u:
@doomspork Not right now. There are interviews going on for that profile. If we don't find someone, then we'll update the link

PR #333, opened 2018-07-12
Update case on solarisBank name

ggpasqualino:


doomspork:
Thank you @ggpasqualino 👍

PR #345, opened 2018-07-24
remove Emagroup

goofansu:
I left the company and it stopped using Elixir.


doomspork:
Bummer!  Thanks for updating though @goofansu

PR #350, opened 2018-07-25
fix dup Marketplace/Marketplaces industry typo

madlep:
At the moment there is a `Marketplace` industry, and `Marketplaces` industry listed.

marketplace marketplaces

Fixed by setting Expert360's industry to `Marketplaces` - which seems to be the convention before it was added.


madlep:
Submitting fix for this one, as I'm wanting to add my company, which fits in `Marketplaces` industry, and figured I'd fix that one first 😃

Issue #296, opened 2018-05-16
Filter/search on location?


Tuxified:
It's currently possible to browse by industry, but I think it might also make sense to add location as a filter so folks can see which companies can be found in their country/neighbourhood? What do you folks think, would that be a good feature?

Nezteb:
I came here to post this same issue. Even just having a page that lists all of the unique locations in the source YAML file and which companies are in those locations would be good.

doomspork:
This is a great idea.  Let's talk it over.  What would we want the final product to look like?

doomspork:
We've updated the site to be powered by Phoenix, I think this issue, and #354, should be easier to tackle now. Happy to discuss this further 👍

Nezteb:
I'll come up with a mockup in the next few days! Also check to see how difficult it would be to implement myself. 😄

doomspork:
@Nezteb I'm happy to brainstorm some approaches with you and do some pair-programming if our schedules overlap 😁

Throwing out some ideas:

We add PostGIS to the DB and a lat/lon for every business then plot them on a map.

doomspork:
Conversation for implementation can be found on #494

PR #354, opened 2018-07-25
[WIP] Locations

lackac:
Related to #296.

This is a first attempt at Locations in an effort to spark the
discussion. It is not expected to be merged back. It's also based on my
other two PRs so I expect this to be rebased on `master` when those are
merged back.

This is a naive approach of just accepting what's been given as location.
Clearly this needs more work and some decisions, but it helps with
revealing the what needs to be done to make this usable.

The way I see it:

- [ ] normalising locations (unify different spelling programatically or by
enforcing it at contribution time)
- [ ] decide on a common way including multiple locations (e.g. "London;
Berlin"), document it in CONTRIBUTING and make existing data conform to it
- [ ] generate higher level groupings by parsing country/area from location
(e.g. UK, India, USA, or maybe even Europe)

### DIFF
--- a/src/_includes/_nav.html
+++ b/src/_includes/_nav.html
@@ -12,22 +12,8 @@
     </div> <!-- /navbarbrand -->
     <div class="navbar-menu" id="navMain">
       <div class="navbar-start">
-        <div class="navbar-item has-dropdown is-hoverable">
-          <a class="navbar-link is-active">Industries</a>
-          <div id="moreDropdown" class="navbar-dropdown">
-            <div class="columns">
-              {% assign per_column = site.data.industries | size | divided_by: 3.0 | ceil %}
-              {% for col in (0..2) %}
-                {% assign offset = per_column | times: col %}
-              <div class="column">
-                {% for industry in site.data.industries offset: offset, limit: per_column %}
-                <a class="navbar-item" href="/industries/{{ industry['path'] }}"><span class="has-text-weight-bold">{{ industry['name'] }}</span></a>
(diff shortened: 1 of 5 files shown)


OmanF:
I think a way at getting country normalization would be to *stop* allowing pull-request for the `.yml` file including the list of companies, opting instead, on adding to the website a simple form:
1. Company name.
2. Website.
3. Github (if they have some public account).
4. A short description of the company.
5. A drop-down menu of world countries (like the one you have on just about signup site, where you need to specify your country of residence).

Of course, all but the company name should be optional (yes, even the company's website, though I'd be wary of such an entry, as a user).

**This form *and only this form* will be allowed to merge into the `.yml` file.** (i.e. via a webhook, or whatnot, that rejects all PRs changing that file, unless they came from the form).

As an added bonus, implementing this way of entering new companies gives you:
1. Consistency between all newly added comps (don't know what to do about existing ones... use some Levenshtein distance algorithm to try and guess?! Come to think about it, it **will** work: the world has only so many countries, and their names are, usually, very unique).
2. Easy filtering on the country: just select the country you want from *the same drop-down list* and the server pulls all the comps with that value in their "Country" field. (What would that be? SQL? Pattern matching? Hashtable... I honestly don't know which of these has the best big-O performance for *fetching* data).

Just my 2 (euro)cents.

doomspork:
Closing this for now because as of #438, we're using Phoenix. I think this would be easier with Phoenix, would you be interested in continuing this? I'm happy to create an issue to continue the conversation.

PR #371, opened 2018-08-28
Making some changes in the industries list

OmanF:
The first step toward consolidating the list of available industries to filter on.
Not saying this is the best way to go, but it will at least be less of an eyesore until the dust settles.

### DIFF
--- a/src/_data/companies.yml
+++ b/src/_data/companies.yml
@@ -15,13 +15,13 @@
   location: "Guildford, Surrey, United Kingdom"
 
 361° DIGITAL:
-  industry: Technology Consulting
+  industry: BI, Consultancy, Enterprise, Networking
   location: Stuttgart, Germany
   url: https://361.de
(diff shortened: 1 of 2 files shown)


doomspork:
@michalmuskala, @lackac, among others:  any thoughts on this approach?

lackac:
I skimmed through the changes and I don't particularly like this approach. It conflates arguably unrelated industries. For example: "BI, Consultancy, Enterprise, Networking".

I think it would be better to collect some higher level categories and maybe group industries under that. For example, one way to categorise would be to make it clear whether a company is a "product company" or a "consultancy" or some other sector that happens to use Elixir (e.g. kloeckner.i).

Maybe we could look into some taxonomies others use. There might be even a standardised one. However, I do not have the time to look into this more deeply at the moment and will accept the choice made by the maintainers.

doomspork:
Closing this as we've now ditched Jekyll in favor of a Phoenix application. If we still want to pursue this, I'd love to see it ported to Phoenix 👍

PR #389, opened 2018-09-25
Update Telnyx's industry

rranelli:
The "Telecommunications" industry makes more sense for Telnyx than "Communication";


doomspork:
Thank you @rranelli 👍

PR #408, opened 2018-11-01
Remove Kadabra

carlodicelico:
Positions filled quite awhile ago, apologies for the delay! Also, we have rebranded. I'll add us back later once things settle down a bit. 🤓


doomspork:
Thank you @carlodicelico 😁

PR #429, opened 2019-01-31
Update Podium

dnsbty:
A little different than normal since I was actually updating Podium's current listing, but here's the checklist for good measure.

- [x] Company name
- [x] Industry
- [x] Company location
- [x] Company description
- [x] How Elixir is used
- [x] GitHub / blog links
- [x] Job listings


doomspork:
Thank you @dnsbty!!

PR #437, opened 2019-02-11
update Pharos Production Inc. information

dmytronasyrov:
update Pharos Production Inc. information


doomspork:
Thank you @dmytronasyrov!

PR #444, opened 2019-02-25
simple link to show companies per industry

gemantzu:
Next step is to add a list of industries and make it possible to browse for each one of them. This is just a simple way to view the other options on a specific industry, by clicking on it in a company card.

I removed most of the query in Industries.all, because it was kinda in the way (if there was not any companies associated with an industry, it would not show, even in the create form).

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -41,6 +41,12 @@ defmodule Companies.Companies do
     |> order_by([_c, j], desc: j.inserted_at)
   end
 
+  defp predicates(query, %{"industry" => industry_id}) do
+    query
+    |> where([c], c.industry_id == ^industry_id)
+    |> order_by(asc: :name)
+  end
+
   defp predicates(query, _) do
     order_by(query, [c, _i, _j], desc: c.inserted_at)
   end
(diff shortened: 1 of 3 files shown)


doomspork:
@gemantzu what does this provide exactly? Can we navigate to this any other way? If not, can we hold off until we have all of that functionality built?

I'd love to hear @burden or @tajchumber's thoughts on how we can handle the different filters for companies: name, location, industry, hiring, ...

gemantzu:
I don't mind waiting, I just wanted to bring back some of the old functionality. We could have another tab (Per industry) after the letters which would list all industries that have companies, with a link leading to what I have done here.

tajchumber:
how about adding a search functionality we can take inspiration from stackoverflow jobs? (see image below)
screenshot 2019-02-26 at 21 21 24

We could add filters to view companies by industry, location and if they are hiring or not.

I haven't looked at the code yet, I need a bit of time to look at what it does

doomspork:
@tajchumber that would be amazing 🎉

tajchumber:
I will have a look into it this weekend on how to implement this in the best way!

doomspork:
@tajchumber any updates on filter improvements?

@gemantzu closing this for now until we figure out a path towards better industry support.

PR #447, opened 2019-02-26
Add scheme for url

### LINKED ISSUE #446: Urls when http / https is missing
When http is missing, it converts the url to be a part of the website ("https://elixir-companies.com/url").

Edit: not just the website url, we need to force users to enter the correct protocol for their website, so we do not have to guess which one works.

gonglexin:
Fix #446

### DIFF
--- a/lib/companies_web/views/company_view.ex
+++ b/lib/companies_web/views/company_view.ex
@@ -16,6 +16,10 @@ defmodule CompaniesWeb.CompanyView do
     |> String.replace("www.", "")
   end
 
+  def url_with_scheme("http://" <> _ = url), do: url
+  def url_with_scheme("https://" <> _ = url), do: url
+  def url_with_scheme(url, scheme \\ "http"), do: "#{scheme}://#{url}"
+
   def render("url.html", %{url: nil}), do: ""
   def render("blog.html", %{blog: nil}), do: ""
   def render("github.html", %{github: nil}), do: ""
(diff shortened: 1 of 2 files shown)


lib/companies_web/views/company_view.ex
```diff
@@ -16,6 +16,10 @@ defmodule CompaniesWeb.CompanyView do
     |> String.replace("www.", "")
   end
 
+  def url_with_scheme("http://" <> _ = url), do: url
+  def url_with_scheme("https://" <> _ = url), do: url
+  def url_with_scheme(url, scheme \\ "http"), do: "#{scheme}://#{url}"
```
doomspork: I think we want the default to be `https`, what do you think?

doomspork:
Thank you for the PR @gonglexin! One comment 👍

lib/companies_web/views/company_view.ex
gemantzu: Even one or the other might lead to issues (like a url not been set properly to redirect to one or the other), we maybe want to do something more clever on the long run. It's fine for now in https I guess

doomspork: I can look into the cloudflare setup to ensure http goes to https.

gemantzu: But that's the thing, one might have setup their domain to only use one of them (I know, I have done this shit already :blush)

tajchumber: how about trying to retrieve the endpoint from here?
https://github.com/doomspork/elixir-companies/blob/2872b679c40b468e7ae4c01e0d6c34deb4bc7545/config/prod.exs#L14

gonglexin: I'm thinking default scheme to https and open another issue for adding scheme required validation for company & job url, what's your opinion?

doomspork: @gonglexin sounds good to me 👍

PR #455, opened 2019-02-26
Trigger a Slack message for any pending change

### LINKED ISSUE #445: Alert maintainers when a change is submitted
Right now we need to check regularly to see if there are pending changes, we should be alerted when these occur. For now we can stick with a Slack webhook 👍

doomspork:
Closes #445


doomspork:
image

Issue #439, opened 2019-02-25
Fix the favicon


doomspork:
The favicon appears to be missing, we should fix that.

doomspork:
@burden / @tajchumber would one of you be able to look at this?

PR #456, opened 2019-02-26
fix favicon,  #439

tajchumber:
fix for https://github.com/doomspork/elixir-companies/issues/439


doomspork:
Thank you @tajchumber!

PR #457, opened 2019-02-26
Update application name and namespace to match repo

doomspork:
Big change I'd love to get another set or two of 👀 on this to make sure I didn't miss anything or over apply my replacement.


doomspork:
Now that I've slept on this I'm curious: do we want to do this @gemantzu / @tajchumber? The site _is_ elixir-companies.com but the underlying application can just be `Companies`.

gemantzu:
That's a question I wanted to ask. I think we must leave it as companies, since this can be used by others for different languages. We can extract this as a package as well I believe, and let others use and expand on it as well.

PR #459, opened 2019-02-27
Update README.md

gemantzu:
Just an FYI if someone resets the db in their local machine.


tajchumber:
shouldn't the app clear the cookies if no user is found in db? :)

PR #462, opened 2019-02-27
WIP: edit & delete functionality (needs testing)

gemantzu:
Please test this and leave comments.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -73,3 +73,5 @@ config :companies, Companies.Repo,
   database: "companies_dev",
   hostname: "localhost",
   pool_size: 10
+
+config :companies, :notify_slack, false
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -126,9 +126,7 @@ defmodule Companies.Companies do
 
   """
   def delete_company(%Company{} = company, user) do
-    company
-    |> Company.changeset(%{})
-    |> PendingChanges.create(:delete, user)
+    PendingChanges.create(%{data: company, params: %{"id" => company.id, "delete" => true}}, :delete, user)
   end
 
   @doc """
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -107,10 +107,10 @@ defmodule Companies.Companies do
   {:error, %Ecto.Changeset{}}
 
   """
-  def update_company(%Company{} = company, attrs) do
+  def update_company(%Company{} = company, attrs, user) do
     company
     |> Company.changeset(attrs)
-    |> Repo.update()
+    |> PendingChanges.create(:update, user)
   end
 
   @doc """
@@ -125,8 +125,10 @@ defmodule Companies.Companies do
   {:error, %Ecto.Changeset{}}
 
   """
-  def delete_company(%Company{} = company) do
-    Repo.delete(company)
+  def delete_company(%Company{} = company, user) do
+    company
+    |> Company.changeset(%{})
+    |> PendingChanges.create(:delete, user)
   end
 
   @doc """
--- a/lib/companies/jobs.ex
+++ b/lib/companies/jobs.ex
@@ -87,9 +87,7 @@ defmodule Companies.Jobs do
 
   """
   def delete_job(%Job{} = job, user) do
-    job
-    |> Job.changeset(%{})
-    |> PendingChanges.create(:delete, user)
+    PendingChanges.create(%{data: job, params: %{"id" => job.id, "delete" => true}}, :delete, user)
   end
 
   @doc """
--- a/lib/companies/jobs.ex
+++ b/lib/companies/jobs.ex
@@ -68,10 +68,10 @@ defmodule Companies.Jobs do
       {:error, %Ecto.Changeset{}}
 
   """
-  def update_job(%Job{} = job, attrs) do
+  def update_job(%Job{} = job, attrs, user) do
     job
     |> Job.changeset(attrs)
-    |> Repo.update()
+    |> PendingChanges.create(:update, user)
   end
 
   @doc """
@@ -86,8 +86,10 @@ defmodule Companies.Jobs do
       {:error, %Ecto.Changeset{}}
 
   """
-  def delete_job(%Job{} = job) do
-    Repo.delete(job)
+  def delete_job(%Job{} = job, user) do
+    job
+    |> Job.changeset(%{})
+    |> PendingChanges.create(:delete, user)
   end
 
   @doc """
--- a/lib/companies/pending_changes.ex
+++ b/lib/companies/pending_changes.ex
@@ -41,6 +41,7 @@ defmodule Companies.PendingChanges do
 
     {:ok, :rejected}
   end
+
   def act(change_id, "true") do
     change_id
     |> get!()
@@ -52,7 +53,6 @@ defmodule Companies.PendingChanges do
          record <- get_record(module, changes),
          changeset <- apply_changeset(module, action, record, changes),
          {:ok, _changes} <- apply_changes(action, changeset) do
-
       change
       |> PendingChange.changeset(%{approved: true})
       |> Repo.update()
@@ -65,6 +65,7 @@ defmodule Companies.PendingChanges do
   defp apply_changeset(_module, "delete", record, _changes) do
     record
   end
+
   defp apply_changeset(module, _action, record, changes) do
     module.changeset(record, changes)
   end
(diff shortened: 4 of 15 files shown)


lib/companies/pending_changes.ex
```diff
@@ -75,6 +76,18 @@ defmodule Companies.PendingChanges do
...
+  defp current_values(%{changes: %{"id" => id}, resource: "job"}) do
+    {%{id: company_id}, current_values} =
+      Job
+      |> Repo.get(id)
+      |> Repo.preload([:company])
+      |> drop_ecto_fields()
+      |> drop_nulls()
+      |> Map.pop(:company)
+
+    Map.put(current_values, :company_id, company_id)
```
doomspork: If you drop the `preload` you'll be left with `company_id`, won't you?

gemantzu: Damn, that's true. I guess when one does a pr @ 2 in the morning, this happens :)

lib/companies/jobs.ex
```diff
@@ -86,8 +86,10 @@ defmodule Companies.Jobs do
       {:error, %Ecto.Changeset{}}
 
   """
-  def delete_job(%Job{} = job) do
-    Repo.delete(job)
+  def delete_job(%Job{} = job, user) do
+    job
+    |> Job.changeset(%{})
```
doomspork: We shouldn't need to create a changeset just to delete.

gemantzu: This is one of the things i was working on today, but didn't finish until i left home. The whole thing needs some rework, plus there are some holes we have in the original branch as well. I am going to work on them tomorrow and during the weekend.

gemantzu:
Latest commit handles #464 as well.

config/dev.exs
```diff
@@ -73,3 +73,5 @@ config :companies, Companies.Repo,
   database: "companies_dev",
   hostname: "localhost",
   pool_size: 10
+
+config :companies, :notify_slack, false
```
doomspork: I think we should solve this a different way.

lib/companies/slack.ex
```diff
@@ -2,7 +2,23 @@ defmodule Companies.Slack do
   use HTTPoison.Base
 
   def notify(%{action: action, resource: resource, user: %{nickname: nickname}}) do
-    message = %{
+    case notify_slack() do
```
doomspork: I think we should configure which module to use rather than this bool. Here we lose our notifications in dev which is not useful.

```elixir
defmodule Notify.Console do
  use HTTPoison.Base

  require Logger

  def notify(type, _resource), do: Logger.info("NOTIFICATION — #{type}")
end
```

lib/companies/companies.ex
```diff
@@ -125,8 +125,8 @@ defmodule Companies.Companies do
   {:error, %Ecto.Changeset{}}
 
   """
-  def delete_company(%Company{} = company) do
-    Repo.delete(company)
+  def delete_company(%Company{} = company, user) do
+    PendingChanges.create(%{data: company, params: %{"id" => company.id, "delete" => true}}, :delete, user)
```
doomspork: `"delete" => true` seems redundant when passing `:delete`?

lib/companies/jobs.ex
```diff
@@ -86,8 +86,8 @@ defmodule Companies.Jobs do
       {:error, %Ecto.Changeset{}}
 
   """
-  def delete_job(%Job{} = job) do
-    Repo.delete(job)
+  def delete_job(%Job{} = job, user) do
+    PendingChanges.create(%{data: job, params: %{"id" => job.id, "delete" => true}}, :delete, user)
```
doomspork: Similar thought RE: `delete => true`

lib/companies/pending_changes.ex
```diff
@@ -9,32 +9,15 @@ defmodule Companies.PendingChanges do
   alias Companies.Schema.{Company, Job, Industry, PendingChange}
 
   def all do
-    query =
-      from p in PendingChange,
-        select: p,
-        where: p.approved == false
+    query = from p in PendingChange, where: p.approved == false
```
doomspork: What was the thinking behind this change?

lib/companies/pending_changes.ex
```diff
@@ -9,32 +9,15 @@ defmodule Companies.PendingChanges do
...
-         changeset <- changeset(module, changes),
-         {:ok, _changes} <- apply_changes(action, changeset) do
-      change
-      |> PendingChange.changeset(%{approved: true})
-      |> Repo.update()
-    else
-      nil -> {:error, "change not found"}
-      {:error, changeset} -> {:error, changeset}
-    end
+  def get!(id) do
```
doomspork: Do we really need to define this? We already have `get/1` and this isn't going to widely consume. I think a smaller API makes testing things easier and reduces the number of potential code paths

lib/companies/pending_changes.ex
```diff
@@ -47,62 +30,66 @@ defmodule Companies.PendingChanges do
     end
   end
 
-  defp apply_changes("create", changeset), do: Repo.insert(changeset)
-  defp apply_changes("update", changeset), do: Repo.update(changeset)
+  def create(changeset, action, user, note \\ "")
```
doomspork: Why did you move these? They were in alphabetical order before.

lib/companies/pending_changes.ex
```diff
@@ -47,62 +30,66 @@ defmodule Companies.PendingChanges do
...
 
-  defp changeset(module, changes) do
-    record =
-      if Map.has_key?(changes, "id") do
-        id = Map.get(changes, "id")
-        Repo.get(module, id)
-      else
-        struct(module, %{})
-      end
+  def act(change_id, "false") do
```
doomspork: I feel like this `act/2` functions is entirely unnecessary. Why can't we pass in a boolean to the existing method?

lib/companies/pending_changes.ex
```diff
@@ -47,62 +30,66 @@ defmodule Companies.PendingChanges do
...
+    PendingChange
+    |> Repo.get!(change_id)
+    |> Repo.delete!()
```
doomspork: We're not deleting anything. We just mark it as rejected. We want the record.

lib/companies/pending_changes.ex
```diff
@@ -138,9 +142,17 @@ defmodule Companies.PendingChanges do
...
   end
 
+  defp apply_changes("create", changeset), do: Repo.insert(changeset)
+  defp apply_changes("update", changeset), do: Repo.update(changeset)
+  defp apply_changes("delete", record), do: Repo.delete(record)
+
   defp resource_module("company"), do: Company
   defp resource_module("industry"), do: Industry
   defp resource_module("job"), do: Job
+  defp resource_module(_), do: {:error, :module_not_recognized}
```
doomspork: This should never happen, we don't need to worry about it. What brought you here?

lib/companies/pending_changes.ex
```diff
@@ -138,9 +142,17 @@ defmodule Companies.PendingChanges do
...
+
+  defp drop_relations(job = %Job{}), do: Map.drop(job, [:company])
```
doomspork: I don't think we need these methods potentially but if we do, they can probably be consolidated into one.  `Map.drop/2` can handle keys that don't exist.

lib/companies/pending_changes.ex
gemantzu: get/1 is doing some other things. Granted, this should be a private method and not parted of the API.

lib/companies/pending_changes.ex
gemantzu: I found the select redundant, so I removed it.

lib/companies/pending_changes.ex
gemantzu: Sorry, I didn't notice that you had them in alphabetical order and I moved them around in a way that made sense to me (and, I thought it might be in the longs run). I moved the api methods on top and the private below, moving the easy to deal with methods on the far bottom.

lib/companies/pending_changes.ex
gemantzu: It is passing by the value from the buttons in the form (accept => true, reject => false)

lib/companies/pending_changes.ex
gemantzu: Oh ok, nice. I will fix this.

doomspork:
#466 includes this code 👍

PR #470, opened 2019-03-04
Ensure distinct companies. Minimize sub queries

doomspork:
I noticed on the site we display hiring companies as many times as they have openings (ie 2 jobs == 2 cards). While investigating this I also noticed while we had `preload` we were still doing a query each for companies, industries, and jobs.

This PR fixes both of those.

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -22,8 +22,11 @@ defmodule Companies.Companies do
 
     (c in Company)
     |> from()
+    |> join(:inner, [c], i in assoc(c, :industry))
+    |> join(:left, [c, _i], j in assoc(c, :jobs))
     |> predicates(params)
-    |> preload([:industry, :jobs])
+    |> distinct([_c, _i, _j], true)
+    |> preload([_c, i, j], industry: i, jobs: j)
     |> Repo.paginate(page: page, page_size: 8)
   end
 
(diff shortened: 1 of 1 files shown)


lib/companies/companies.ex
```diff
@@ -22,8 +22,11 @@ defmodule Companies.Companies do
 
     (c in Company)
     |> from()
+    |> join(:inner, [c], i in assoc(c, :industry))
+    |> join(:left, [c, _i], j in assoc(c, :jobs))
     |> predicates(params)
-    |> preload([:industry, :jobs])
+    |> distinct([_c, _i, _j], true)
+    |> preload([_c, i, j], industry: i, jobs: j)
```
doomspork: So we can preload these specific joins, we need to add both `industry` and `jobs` joins to our query here. Since there may be no jobs, we go with a left join.

lib/companies/companies.ex
```diff
@@ -22,8 +22,11 @@ defmodule Companies.Companies do
 
     (c in Company)
     |> from()
+    |> join(:inner, [c], i in assoc(c, :industry))
+    |> join(:left, [c, _i], j in assoc(c, :jobs))
     |> predicates(params)
-    |> preload([:industry, :jobs])
+    |> distinct([_c, _i, _j], true)
```
doomspork: We want distinct companies, always.

lib/companies/companies.ex
```diff
@@ -134,8 +137,9 @@ defmodule Companies.Companies do
 
   defp predicates(query, %{"type" => "hiring"}) do
     query
-    |> join(:inner, [c], j in assoc(c, :jobs))
-    |> order_by([_c, j], desc: j.inserted_at)
+    |> exclude(:left_join)
```
doomspork: Since we're going to apply an inner join now, let's remove the left join applied in the base query build up.

PR #473, opened 2019-03-04
Remove now-obsolete contributing guide

brain-geek:
At the moment `CONTRIBUTING.md` still talks about adding companies manually, which is no longer the case. It would be simpler to remove it and add the proper one later.


doomspork:
Good idea 👍

PR #478, opened 2019-03-05
Travis should run our tests

doomspork:


gemantzu:
@tajchumber any idea why this fails to run?

tajchumber:
the error from the travic ci logs is
```elixir
17:23:28.806 [error] GenServer #PID<0.306.0> terminating
** (Postgrex.Error) FATAL 28P01 (invalid_password) password authentication failed for user "postgres"
    (db_connection) lib/db_connection/connection.ex:84: DBConnection.Connection.connect/2
    (connection) lib/connection.ex:622: Connection.enter_connect/5
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: nil
** (Mix) The database for Companies.Repo couldn't be dropped: killed
The command "mix test" exited with 1.
```

it seems like the database is not setup properly. Can you try following the docs for setting up the db here: https://docs.travis-ci.com/user/database-setup/#postgresql

gemantzu:
I did, I tried multiple things. The reason I got here, is that at some point tests were passing on my machine and @doomspork 's and not on CI, with an error on postgres query. We bumped the pg version up to test if there was actually something that changed due to version updates. Now I feel that we should just go back to 9.4+ (it is needed for json blob field) and find what bug we have on our code (because at a later point, tests were failing again on my machine, without changing anything on that query).

gemantzu:
OK, I finally made it work. I rebased the branch to master (stupid me, why would the tests pass when the current branch is not based on the fixed master?), bumped postgres to 9.5 (needed for ON CONFCLICT) and rolled back any unneeded changes. I also added a note on README for devs that they should have 9.5+. @doomspork check the changes, and if you deem it necessary do a force-push to remove the clutter.

tajchumber:
all tests pass now, we can merge this into master!

PR #481, opened 2019-03-06
Redirect hiring

### LINKED ISSUE #476: /hiring from google - Route exists but no method in controller
AppSingal shows this:
`function CompaniesWeb.CompanyController.hiring/2 is undefined or private`
which is possible if you come from google, where we have /hiring as a path (probably some seo left from the old website). This path exists in our router, but internally we have changed this functionality to work through `browse?type=hiring`.

zacck:
Fixes #476 

Changes Proposed 
- [x] redirect `/hiring` route to `/browse?type=hiring`

Reservations 
Built URL by hand
Didnt build any tests (cheeky)

### DIFF
--- a/lib/companies_web/controllers/redirect_plug.ex
+++ b/lib/companies_web/controllers/redirect_plug.ex
@@ -1,0 +1,16 @@
+defmodule CompaniesWeb.Redirect do
+  @moduledoc """
+  A plug used to handle redirects by the `CompaniesWeb.Router`
+  """
+  import Plug.Conn
+
+  @spec init(Keyword.t) :: Keyword.t
+  def init(opts), do: opts
+
+  @spec call(Plug.Conn.t, Keyword.t) :: Plug.Conn.t
+  def call(conn, opts) do
+    conn
+    |> Phoenix.Controller.redirect(opts)
+    |> halt()
+  end
+end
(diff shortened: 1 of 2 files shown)


gemantzu:
Wow, I haven't seen this approach before, I would typically do a simple redirect on a controller action, care to explain this to me a bit?

zacck:
@gemantzu  oh really I applied this thinking it would be the standard approach since we don't have a matching controller action, So what I do here is I pass the options from the router to my plug, the important bit is the `to` key which accepts a path and the send the user along to that given url

You can see the relevant docs here 
https://hexdocs.pm/phoenix/Phoenix.Controller.html#redirect/2 ... I can switch it up to  standard redirect if you feel this is overkill?

gemantzu:
The issue I came up with while testing your branch is that you do not have a :to key set anywhere, and as a result, Phoenix.redirect/2 fails. Was it working on your machine? You probably wanted to do it in the router, instead of using `get "/hiring", Redirect, "browse?type=hiring"` => `get "/hiring", Redirect, to: "/browse?type=hiring"`

zacck:
@gemantzu  aah thats was my last change I didn't add it to my amend, good catch

gemantzu:
You have to add a leading "/" in the path, otherwise it's not considered as one. :)

zacck:
@gemantzu your diligence is not human!

doomspork:
Thank you @zacck and @gemantzu 💜

PR #483, opened 2019-03-06
Add automatic periodic check of URLs in jobs

brain-geek:


lib/companies/jobs.ex
```diff
@@ -86,6 +86,10 @@ defmodule Companies.Jobs do
       {:error, %Ecto.Changeset{}}
 
   """
+  def delete(%Job{} = job) do
+    Repo.delete(job)
+  end
+
```
gemantzu: Would it be a good idea to add a `checker` user or something uniquely named which can indicate which one made the change, and as a result leave the api intact?

brain-geek: I guess it would be nice idea to just implement soft-delete all over the app, this would simplify everything. Not so sure about how to store delete reasons. String i.e. "deleted by automatic url checker", "deleted by github user ____"?

brain-geek:
@notactuallypagemcconnell yeah, sure, the only problem I don't like in `send_after` approach is that I have no idea how to test the recurring part.

notactuallypagemcconnell:
@brain-geek yes I will push up a commit with an example sometime today

notactuallypagemcconnell:
@brain-geek heres an example, I just will paste here since this branch is on your fork and I'm lazy.

```
# lib/companies/periodic.ex
defmodule Companies.Periodic do
  use GenServer
  require Logger

  def start_link({_mod, _func, _argument_list, _interval} = state) do
    GenServer.start_link(__MODULE__, state)
  end

  @impl true
  def init({_mod, _func, _argument_list, interval} = state) do
    schedule_work(interval)

    {:ok, state}
  end

  @impl true
  def handle_info(:work, {mod, func, argument_list, interval} = state) do
    apply(mod, func, argument_list)
    schedule_work(interval)

    {:noreply, state}
  end

  defp schedule_work(interval) do
    Process.send_after(self(), :work, interval)
  end
end
```
Code to run off `{mod, func, args, interval}` where func is applied to mod using args in interval

```
# test/companies/periodic_test.exs
defmodule Companies.PeriodicTest do
  use ExUnit.Case

  import ExUnit.CaptureIO

  require Logger

  test "it can run a given module's given function with the given arguments in the given interval" do
    assert capture_io(fn ->
             {:ok, _pid} = Companies.Periodic.start_link({IO, :puts, ["hi"], 1})
             :timer.sleep(5)
           end) =~ "hi\nhi"
  end
end
```

This test proves that it runs the task twice and has minimal overhead (3 milliseconds spent sleeping) but proves the premise is sound and tests it at an OTP level. I think it works well enough. We use some code similar to this (but with some more sophistication added) at work.

brain-geek:
@notactuallypagemcconnell Thanks for the snippet, that's a clever way to test it! Applied.

lib/companies/job_checker.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Companies.JobChecker do
+  alias Companies.Jobs
+
+  def check_job(job) do
```
doomspork: If possible, I'd like to avoid pipelines in situations like this. What do you think about capturing `url`? I think we can also eliminate the duplicate `nil` paths with a `_` match:

```suggestion
def check_job(%{url: url} = job) do
  with {:ok, %{status_code: 404}} <- HTTPoison.head(url) do
    Jobs.delete(job)
  else
    _ -> nil
  end
end
```

```suggestion
def check_job(%{url: url} = job) do
  case HTTPoison.head(url) do
    {:ok, %{status_code: 404}} -> Jobs.delete(job)
    _ -> nil
  end
end
```

lib/companies/job_checker.ex
```diff
@@ -0,0 +1,20 @@
...
+      {:ok, %HTTPoison.Response{}} ->
+        nil
+
+      {:error, _} ->
+        nil
+    end
+  end
+
+  def check_all do
+    Jobs.all() |> Enum.each(&check_job(&1))
```
doomspork: Could we change this to:

```suggestion
Enum.each(Jobs.all(), &check_job(&1))
```

https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines

lib/companies/jobs.ex
doomspork: We could devise a way to leverage the PendingChange tracking (which has `note` and an associated `user`) for this with an auto-approval for changes submitted via this checker (and admins)?

brain-geek:
@doomspork maybe, but then we need to mark this automatic user somehow.

lib/companies/jobs.ex
brain-geek: Yep, and also add a soft-delete from #488 as a safety net.

lib/companies/scheduler.ex
```diff
@@ -0,0 +1,26 @@
...
+  @impl true
+  def handle_info(:work, {mod, func, argument_list, interval} = state) do
+    apply(mod, func, argument_list)
+    schedule_work(interval)
+
+    {:noreply, state}
+  end
+
+  defp schedule_work(interval) do
+    Process.send_after(self(), :work, interval)
```
Nitrino: In this implementation, each deployment (with restart) will shift the task execution, thus if it is often deployed, a task can be performed very rarely. 
Although I think this is not critical for us, it will be possible to add task execution immediately after the start of the gen_server.

brain-geek: Having the task run on start would be bad if app is crashing, for example. OTP will continue restarting it and we'll create load on other people's servers.

Having this delayed is intentional to make sure we make as least amount of requests as possible.

Also, it would be an interesting case if we had a problem in this checking jobs code. Then this OTP process will fail, restart, and scheduled once again. In case of immediate run - we'll be accessing few first jobs immediately, and again and again.

doomspork: We could potentially leverage a log to prevent excess execution based on last entry timestamp.

Nitrino: I meant not executing checker code directly in gen_server init, and once sending an asynchronous message to yourself, same as by timeout.

But if the execution frequency is not important, then let's just skip it.

config/config.exs
```diff
@@ -36,6 +36,11 @@ config :companies,
     ]
   }
 
+config :companies, :jobs_url_checker,
+  enabled: true,
+  # 3 hours
+  interval: 10_800
```
doomspork: I realistically I think we could make this much higher, closer to 12 hours, maybe even a whole day. We don't need to update the listings in real-time.

lib/companies/job_checker.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Companies.JobChecker do
+  alias Companies.Jobs
+
+def check_job(%{url: url} = job) do
+  case HTTPoison.head(url) do
+    {:ok, %{status_code: 404}} -> Jobs.delete(job)
+    _ -> nil
+  end
+end
+    case job.url |> HTTPoison.head() do
```
doomspork: I think something happened here.

config/config.exs
doomspork: Sounds good to me 👍

lib/companies/job_checker.ex
brain-geek: yup, github happened with their "apply changes" feature: https://github.com/doomspork/elixir-companies/pull/483/commits/ea2efe86411d8cae1ec49eb423b8c4039ccf1d14

gemantzu: There is still an issue here, can you change it and push it again to check? thnx :)

lccezinha:
Recently I found a blog post that may help with the implementation:

- https://blog.smartlogic.io/genserver-nightly-task/
- https://github.com/oestrich/grapevine/pull/78/files

@gemantzu @brain-geek

brain-geek:
@lccezinha Thanks for the suggestion, reworked it according to your suggestion.

brain-geek:
@doomspork Could you check out the new version? It's now creates pending changes request instead of deleting something, so it should be more reliable. 😄

lib/companies/job_checker.ex
```diff
@@ -0,0 +1,25 @@
...
+  @moduledoc """
+  The module to handle automated PendingChange creation for the no longer present jobs.
+  """
+
+  alias Companies.{Jobs, PendingChanges}
+
+  def check_job(%{url: url} = job) do
+    case HTTPoison.head(url) do
+      {:ok, %{status_code: 404}} ->
+        user_id = Application.get_env(:companies, :jobs_url_checker)[:user_id]
```
michaelstalker: `Application.get_env(:companies, :jobs_url_checker)[:user_id]` resolves to `System.get_env("JOB_CHECKER_USER_ID")` in a production environment, and `nil` in other environments. Would it be simpler to read from the environment variable here and omit setting `user_id` in the configuration file?
```ex
user_id = System.get_env("JOB_CHECKER_USER_ID")
```
That has a slight potential benefit of allowing you to change the value without restarting the app, although I'm not sure if that will make a practical difference. It also simplifies the application logic a little bit. A minor downside is that a change like this splits the configuration up between the configuration file and the application code. It may also be detrimental to use the `JOB_CHECKER_USER_ID` environment variable in non-production environments if that value could be non-empty, and you wanted to ensure `user_id` was `nil` in those environments.

Just some food for thought :smile:

lib/companies/pending_changes.ex
```diff
@@ -71,6 +71,19 @@ defmodule Companies.PendingChanges do
...
+    id = record.id |> to_string()
+
+    query =
+      from p in PendingChange,
+        where: p.resource == ^resource_name,
+        where: is_nil(p.approved),
+        where: fragment("changes->>'id' = ?", ^id)
+
+    Repo.all(query)
+  end
```
michaelstalker: The use case for this function is to determine if any pending changes exist. The place that calls this function isn't really interested in the actual list of changes. What would you think about converting this to a [predicate function](https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo)?

```ex
def any?(record) do
  # ...

  Repo.exists?(query)
end
```

`Companies.JobChecker.check_job/1` could then simplify its conditional like this:
```ex
if PendingChanges.any?(record) do 
  Jobs.delete(job, %{id: user_id})
end
```

lib/companies/pending_changes.ex
```diff
@@ -71,6 +71,19 @@ defmodule Companies.PendingChanges do
     end
   end
 
+  def get_pending_changes_for(record) do
+    resource_name = struct_to_string(record)
+    id = record.id |> to_string()
```
michaelstalker: Since there's only one pipe here, what would you think about using the conventional function syntax?
```ex
id = to_string(record.id)
```

doomspork: This is the preferred way given the styleguide we try to follow. We may need to update our Credo (or add it) rules for some of these.

Additionally I'd likely capture the `record_id` value via a match and avoid dot notation (I don't know why, I don't much care for it personally).

Issue #480, opened 2019-03-06
Hiring ribbon is on top of company name


gemantzu:
![default](https://user-images.githubusercontent.com/4966172/53865890-e8b31300-3ff8-11e9-9d0f-d9458e170f77.png)

As you can see, hiring label hides a part of the text for company name. Should we change this or is it as intended?
@doomspork / @tajchumber / @burden thoughts?

lccezinha:
In my vision the label is important, but maybe should be in another position to avoid hiding the company name.

What about use Ribbon in this position?

![image](https://user-images.githubusercontent.com/2160392/53885610-ef438980-3ffc-11e9-96bf-cec8704cfd09.png)

doomspork:
I like the proposed angled ribbon @lccezinha!

tajchumber:
just remember that if the company name is very long it can still go under the ribbon!
To make sure it doesn't add some `margin-right` to the company name!

doomspork:
Is there an icon we could use that that would indicate hiring, something that could go next to the name?

lccezinha:
According to the google images ![hiring icon](https://cdn3.iconfinder.com/data/icons/business-solid-3/64/Business_Management_Employment_Job-01-512.png)

tajchumber:
@doomspork I think the ribbon is a better solution as it's more explicit

gemantzu:
:thinking: But that icon could be useful in our developer listing...

lccezinha:
Closed in #484

PR #484, opened 2019-03-07
Rotate hiring ribbon element

lccezinha:
Try to solve the #480 problem.

I'm not so good with CSS and frontend so I try to do my best, any thoughts about it?

hiring-ribbon


doomspork:
Thank you @lccezinha! I'll let @burden or @tajchumber review this one, I'm quite possibly the worst at CSS.

lccezinha:
Hey, @tajchumber how about now?

image

PR #488, opened 2019-03-08
Add soft-deletion to companies, jobs, industries

brain-geek:
The first version, feel free to give comments.

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -24,8 +24,8 @@ defmodule Companies.Companies do
     |> from()
     |> predicates(params)
     |> order_by(^order)
-    |> preload([_c, i, j], industry: i, jobs: j)
     |> where([c, _i, _j], is_nil(c.removed_pending_change_id))
+    |> preload([:industry, :jobs])
     |> Repo.paginate(page: page)
   end
 
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -24,7 +24,8 @@ defmodule Companies.Companies do
     |> from()
     |> predicates(params)
     |> order_by(^order)
-    |> preload([:industry, :jobs])
+    |> preload([_c, i, j], industry: i, jobs: j)
+    |> where([c, _i, _j], is_nil(c.removed_pending_change_id))
     |> Repo.paginate(page: page)
   end
 
@@ -69,9 +70,10 @@ defmodule Companies.Companies do
 
   """
   def count do
-    query = from c in Company, select: count(c.id)
-
-    Repo.one(query)
+    from(c in Company)
+    |> where([c], is_nil(c.removed_pending_change_id))
+    |> select([c], count(c.id))
+    |> Repo.one()
   end
 
   @doc """
@@ -91,8 +93,10 @@ defmodule Companies.Companies do
   def get!(id, opts \\ []) do
     preloads = Keyword.get(opts, :preloads, [])
 
-    Company
+    from(c in Company)
     |> preload(^preloads)
+    |> from()
+    |> where([c], is_nil(c.removed_pending_change_id))
     |> Repo.get!(id)
   end
 
--- a/lib/companies/jobs.ex
+++ b/lib/companies/jobs.ex
@@ -22,6 +22,7 @@ defmodule Companies.Jobs do
 
     (j in Job)
     |> from()
+    |> where([j], is_nil(j.removed_pending_change_id))
     |> predicates(params)
     |> order_by(desc: :updated_at)
     |> preload(:company)
@@ -42,7 +43,12 @@ defmodule Companies.Jobs do
       ** (Ecto.NoResultsError)
 
   """
-  def get!(id), do: Repo.get!(Job, id)
+  def get!(id) do
+    (j in Job)
+    |> from()
+    |> where([j], is_nil(j.removed_pending_change_id))
+    |> Repo.get!(id)
+  end
 
   @doc """
   Submits a job listing for approval.
--- a/lib/companies/jobs.ex
+++ b/lib/companies/jobs.ex
@@ -20,12 +20,16 @@ defmodule Companies.Jobs do
   def all(params \\ %{}) do
     page = Map.get(params, "page", "1")
 
-    (j in Job)
-    |> from()
-    |> where([j], is_nil(j.removed_pending_change_id))
+    query =
+      from j in Job,
+        join: c in assoc(j, :company),
+        where: is_nil(j.removed_pending_change_id),
+        where: is_nil(c.removed_pending_change_id),
+        preload: [company: c],
+        order_by: [desc: :updated_at]
+
+    query
     |> predicates(params)
-    |> order_by(desc: :updated_at)
-    |> preload(:company)
     |> Repo.paginate(page: page)
   end
 
@@ -44,10 +48,16 @@ defmodule Companies.Jobs do
 
   """
   def get!(id) do
-    (j in Job)
-    |> from()
-    |> where([j], is_nil(j.removed_pending_change_id))
-    |> Repo.get!(id)
+    query =
+      from j in Job,
+        join: c in assoc(j, :company),
+        where: is_nil(j.removed_pending_change_id),
+        where: is_nil(c.removed_pending_change_id),
+        where: j.id == ^id,
+        preload: [company: c],
+        order_by: [desc: :updated_at]
+
+    Repo.one!(query)
   end
 
   @doc """
--- a/lib/companies/pending_changes.ex
+++ b/lib/companies/pending_changes.ex
@@ -46,7 +46,7 @@ defmodule Companies.PendingChanges do
     with %{action: action, changes: changes, resource: resource} = change <- Repo.get(PendingChange, change_id),
          module <- resource_module(resource),
          changeset <- changeset(module, changes),
-         {:ok, _changes} <- apply_changes(action, changeset) do
+         {:ok, _changes} <- apply_changes(action, changeset, change_id) do
       update_approval(change, note, reviewer, true)
     else
       nil -> {:error, "change not found"}
@@ -71,9 +71,15 @@ defmodule Companies.PendingChanges do
     end
   end
 
-  defp apply_changes("create", changeset), do: Repo.insert(changeset)
-  defp apply_changes("delete", record), do: Repo.delete(record)
-  defp apply_changes("update", changeset), do: Repo.update(changeset)
+  defp apply_changes("create", changeset, _change_id), do: Repo.insert(changeset)
+
+  defp apply_changes("delete", record, change_id) do
+    record
+    |> Ecto.Changeset.cast(%{removed_pending_change_id: change_id}, [:removed_pending_change_id])
+    |> Repo.update()
+  end
+
+  defp apply_changes("update", changeset, _change_id), do: Repo.update(changeset)
 
   defp changeset(module, changes) do
     record =
--- a/lib/companies/schema/company.ex
+++ b/lib/companies/schema/company.ex
@@ -4,7 +4,7 @@ defmodule Companies.Schema.Company do
   use Ecto.Schema
   import Ecto.Changeset
 
-  alias Companies.Schema.{Industry, Job}
+  alias Companies.Schema.{Industry, Job, PendingChange}
 
   schema "companies" do
     field :blog, :string
@@ -15,7 +15,9 @@ defmodule Companies.Schema.Company do
     field :url, :string
 
     belongs_to :industry, Industry
-    has_many :jobs, Job
+    has_many :jobs, Job, defaults: [removed_pending_change_id: nil]
+    # Optional reference to the change that removed the resource
+    belongs_to :removed_pending_change, PendingChange
 
     timestamps()
   end
--- a/lib/companies/schema/industry.ex
+++ b/lib/companies/schema/industry.ex
@@ -4,12 +4,14 @@ defmodule Companies.Schema.Industry do
   use Ecto.Schema
   import Ecto.Changeset
 
-  alias Companies.Schema.Company
+  alias Companies.Schema.{Company, PendingChange}
 
   schema "industries" do
     field :name, :string
 
     has_many :companies, Company
+    # Optional reference to the change that removed the resource
+    belongs_to :removed_pending_change, PendingChange
 
     timestamps()
   end
--- a/lib/companies/schema/job.ex
+++ b/lib/companies/schema/job.ex
@@ -4,7 +4,7 @@ defmodule Companies.Schema.Job do
   use Ecto.Schema
   import Ecto.Changeset
 
-  alias Companies.Schema.Company
+  alias Companies.Schema.{Company, PendingChange}
 
   schema "jobs" do
     field :title, :string
@@ -12,6 +12,8 @@ defmodule Companies.Schema.Job do
     field :remote, :boolean
 
     belongs_to :company, Company
+    # Optional reference to the change that removed the resource
+    belongs_to :removed_pending_change, PendingChange
 
     timestamps()
   end
--- a/priv/repo/migrations/20190308180634_soft_deletion.exs
+++ b/priv/repo/migrations/20190308180634_soft_deletion.exs
@@ -1,0 +1,23 @@
+defmodule Companies.Repo.Migrations.SoftDeletion do
+  use Ecto.Migration
+
+  def change do
+    alter table(:companies) do
+      add(:removed_pending_change_id, references(:pending_changes))
+    end
+
+    create index(:companies, [:removed_pending_change_id])
+
+    alter table(:jobs) do
+      add(:removed_pending_change_id, references(:pending_changes))
+    end
+
+    create index(:jobs, [:removed_pending_change_id])
+
+    alter table(:industries) do
+      add(:removed_pending_change_id, references(:pending_changes))
+    end
+
+    create index(:industries, [:removed_pending_change_id])
+  end
+end
--- a/test/companies/companies_test.exs
+++ b/test/companies/companies_test.exs
@@ -36,12 +36,6 @@ defmodule Companies.CompaniesTest do
       assert %{page_number: 1, page_size: 16, total_entries: 1, total_pages: 1} = Companies.all()
     end
 
-    test "filters companies by hiring" do
-      [company, _] = insert_list(2, :company)
-      insert(:job, company: company)
-      assert %{total_entries: 1} = Companies.all(%{"type" => "hiring"})
-    end
-
     test "filters companies by text" do
       insert(:company, name: "ZULU")
       insert(:company, name: "BETA")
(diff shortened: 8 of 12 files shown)


brain-geek:
@doomspork @tajchumber @gemantzu @burden Guys, what do you think about this?

In context of #483 I believe having soft-delete will improve traceability a lot.

doomspork:
@brain-geek I will look today! Sorry for the delay. I hope that's okay. What do you say?

brain-geek:
@doomspork Thanks, no worries! 

Just trying to push it out before beginning the next thing. 😄

doomspork:
You're the best @brain-geek! 😁

lib/companies/schema/industry.ex
```diff
@@ -2,13 +2,16 @@ defmodule Companies.Schema.Industry do
...
 
-  alias Companies.Schema.Company
+  alias Companies.Schema.{Company, PendingChange}
 
   schema "industries" do
     field :name, :string
 
     has_many :companies, Company
 
+    field :is_deleted, :boolean
```
doomspork: Since this is a boolean and the `is_` prefix isn't common in Elixir, could we just call this `deleted`?

This feedback would apply to all 3 tables.

Given #493 we might want to avoid changing Industry at all since it'll likely be deleted instead? Whatcha think?

test/companies/companies_test.exs
```diff
@@ -15,6 +15,14 @@ defmodule Companies.CompaniesTest do
       assert %{page_number: 1, page_size: 16, total_entries: 2, total_pages: 1} = Companies.all()
     end
 
+    test "does not retrieve deleted records" do
+      insert(:company, is_deleted: true)
+      insert(:company, is_deleted: false)
```
doomspork: Since `false` is the default, do we need this _and_ line 21? I could see one or the other. If we want > 1 record we can use the `insert_pair` or `insert_list` function.

test/companies/jobs_test.exs
```diff
@@ -14,6 +14,33 @@ defmodule Companies.JobsTest do
       insert_list(2, :job)
       assert 2 == length(Jobs.all())
     end
+
+    test "does not retrieve deleted records" do
+      insert(:job, is_deleted: true)
+      present_job = insert(:job, is_deleted: false)
+      another_present_job = insert(:job)
+
+      # TODO: rewrite me, but how?
```
doomspork:
```elixir
test "does not retrieve deleted records" do
    %{id: deleted_id} = insert(:job, deleted: true)
    insert_list(2, :job)

    refute delete_id in Enum.map(Jobs.all(), &(&1.id))
  end
```

test/companies/jobs_test.exs
```diff
@@ -14,6 +14,33 @@ defmodule Companies.JobsTest do
...
+      job1_id = present_job.id
+      job2_id = another_present_job.id
+      assert [%Job{id: ^job1_id}, %Job{id: ^job2_id}] = Jobs.all()
+    end
+  end
+
+  describe "get!/1" do
+    test "retrieves by id" do
```
doomspork: What do you think of this?

```elixir
test "retrieves by id" do
  %{id: job_id} = insert(:job)

  assert %{id: ^job_id} = Jobs.get!(job_id)
end
```

1. We can, and I personally prefer to, use pattern matching to obtain the job's `id`
1. We can match on plain map. I don't think the test gains or loses anything using `%Job{}`, save keystrokes?

lib/companies/jobs.ex
```diff
@@ -18,7 +18,7 @@ defmodule Companies.Jobs do
 
   """
   def all do
-    Repo.all(Job)
+    Job |> Ecto.Query.where(is_deleted: false) |> Repo.all()
```
doomspork: Could you multiple these pipelines (ln 21, 39) to match the rest of the codebase?

lib/companies/pending_changes.ex
```diff
@@ -55,9 +55,15 @@ defmodule Companies.PendingChanges do
...
   end
 
-  defp apply_changes("create", changeset), do: Repo.insert(changeset)
-  defp apply_changes("delete", record), do: Repo.delete(record)
-  defp apply_changes("update", changeset), do: Repo.update(changeset)
+  defp apply_changes("create", changeset, _change_id), do: Repo.insert(changeset)
+
+  defp apply_changes("delete", record, change_id) do
+    record
+    |> Ecto.Changeset.cast(%{is_deleted: true, deleted_via_id: change_id}, [:is_deleted, :deleted_via_id])
```
doomspork: I don't know that this is necessary. I think we can ditch the `"delete"` action in favor of setting `deleted: true` on the delete route?

lib/companies/companies.ex
```diff
@@ -143,6 +146,7 @@ defmodule Companies.Companies do
   end
 
   defp predicates(query, _) do
-    order_by(query, [c, _i, _j], desc: c.inserted_at)
+    query
+    |> order_by([c, _i, _j], desc: c.inserted_at)
```
doomspork: We don't need a pipeline here since this is single function call.

lib/companies/companies.ex
```diff
@@ -59,7 +60,9 @@ defmodule Companies.Companies do
   ** (Ecto.NoResultsError)
 
   """
-  def get!(id), do: Repo.get!(Company, id)
+  def get!(id) do
+    Company |> Ecto.Query.where(is_deleted: false) |> Repo.get!(id)
```
doomspork: Can we multi-line this pipeline to match the rest of the code?

lib/companies/companies.ex
```diff
@@ -40,7 +41,7 @@ defmodule Companies.Companies do
 
   """
   def count do
-    query = from c in Company, select: count(c.id)
+    query = from c in Company, where: c.is_deleted == false, select: count(c.id)
```
doomspork: We multiline queries elsewhere, mind doing that here?

lib/companies/schema/company.ex
```diff
@@ -13,7 +13,10 @@ defmodule Companies.Schema.Company do
     field :url, :string
 
     belongs_to :industry, Industry
-    has_many :jobs, Job
+    has_many :jobs, Job, defaults: [is_deleted: false]
+
+    field :is_deleted, :boolean
```
doomspork: For all of the schemas can we group the fields together and the associations together? I think that's more often how it's seen and what folks will be most familiar with coming into the code.

test/companies/companies_test.exs
```diff
@@ -28,6 +36,31 @@ defmodule Companies.CompaniesTest do
     end
   end
 
+  describe "get!/1" do
+    test "retrieves by id" do
+      company = insert(:company)
```
doomspork: I left feedback below, on the jobs test, that can be applied here as well.

test/companies/companies_test.exs
```diff
@@ -28,6 +36,31 @@ defmodule Companies.CompaniesTest do
...
+
+      company_id = company.id
+
+      assert %{id: ^company_id} = Companies.get!(company.id)
+    end
+
+    test "does not retrieve deleted record" do
+      company = insert(:company, is_deleted: true)
+
+      assert_raise(Ecto.NoResultsError, fn -> Companies.get!(company.id) end)
```
doomspork: Could you remove the parenthesis from this assert and the one on line 60? We aren't using them elsewhere so let's remain consistent 😁

test/companies/jobs_test.exs
```diff
@@ -14,6 +14,33 @@ defmodule Companies.JobsTest do
...
+
+      job_id = job.id
+
+      assert %Job{id: ^job_id} = Jobs.get!(job.id)
+    end
+
+    test "does not retrieve deleted record" do
+      job = insert(:job, is_deleted: true)
+
+      assert_raise(Ecto.NoResultsError, fn -> Jobs.get!(job.id) end)
```
doomspork: No parenthesis please.

lib/companies/schema/industry.ex
brain-geek: yeah, soft-delete does not fit industry very well. However, I don't like the fact that we still have changesets on it and it can be deleted from UI, so I suggest still having these changes on industry.

test/companies/jobs_test.exs
brain-geek: Nice, Thanks!

test/companies/companies_test.exs
brain-geek: That sounds like a need for reconfiguring `mix format`! 😄

brain-geek:
@doomspork Thanks for the feedback, I will be applying it in the following days.

Nitrino:
Recently, Devon C. Estes has written the article about soft deletion using the PostgreSQL schema. 
http://devonestes.herokuapp.com/soft-delete-with-ecto-3-and-postgres

May use a similar approach, then we will not have to worry about the `where(is_deleted: false)` conditions.
What do you think about it?

doomspork:
Good point @Nitrino! 

When we retweet that article on Elixir School someone had responded suggesting this was not an ideal approach. It might be worth looking into their feedback: https://twitter.com/atomkirk/status/1101566414668869637

brain-geek:
@Nitrino Thanks for the link. I've looked through the article and found it interesting. However, I believe that this project is also intended to be checked out by new Elixir devs, and having some magic inside the postgresql would be hard to follow.

However, it may be better to change `is_deleted` field to something like `removed_pending_change_id` to always have a pending change which led to this deletion.

What do you think?

brain-geek:
@doomspork well, I don't agree with `like UNIQUE of only live records` part in this comment, as you can do unique index with `where` attached, so it's not a big deal.

doomspork:
> However, I believe that this project is also intended to be checked out by new Elixir devs, and having some magic inside the postgresql would be hard to follow.

💯 — very good point. I think this is reason enough not to use it.

> However, it may be better to change is_deleted field to something like removed_pending_change_id to always have a pending change which led to this deletion.

Are you suggesting ditching the boolean `is_deleted` in favor of just checking for the existence of an associated pending change here? I could get behind that 👍

Nitrino:
@brain-geek I agree, the approach with schemas is opaque for new developers.

test/companies/companies_test.exs
brain-geek: Well, because I try to be paranoid in tests - it helps in weird cases like reversed logic.

brain-geek:
@doomspork Please review again. I believe I fixed all the issues. 😄

test/companies/companies_test.exs
doomspork: I'm not sure I follow. I'm just asking why this:

```elixir
insert(:company, is_deleted: false)
insert(:company)
```

Versus this, which is the same (right?):

```elixir
insert_list(2, :company, is_deleted: false)
```

test/support/factories.ex
```diff
@@ -12,6 +12,15 @@ defmodule Companies.Factory do
     }
   end
 
+  def deleted_company_factory do
```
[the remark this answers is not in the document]
doomspork: This is interesting: https://github.com/thoughtbot/ex_machina#full-control-of-factory

gemantzu:
Can I add my two cents, as I am not totally behind doing this this way right now?

My take on this issue would be to add an `active` field on companies, jobs (idk about industries, still trying to figure out what we should do with industries) and let pending_changes do it's job here (make the field active or inactive per request).

Also, to add full trackability on what changes we have, we could change pending_changes to have an record id field (we already have the corresponding table name on it), so we can track ( and probably show? ) what changes and on what date have occured on a record. 

Thoughts?

doomspork:
@gemantzu / @brain-geek where are we with this? Do we want to move ahead with this approach and come back to @gemantzu's comments or look at incorporating that now?

Happy to help with the work 😁

priv/repo/migrations/20190308180634_soft_deletion.exs
```diff
@@ -0,0 +1,23 @@
+defmodule Companies.Repo.Migrations.SoftDeletion do
+  use Ecto.Migration
+
+  def change do
+    alter table(:companies) do
+      add(:removed_pending_change_id, references(:pending_changes))
+    end
+
+    create index(:companies, [:removed_pending_change_id])
```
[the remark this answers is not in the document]
brain-geek: We will be having searches by 'not deleted' all over the site. And postgresql can merge indexes: `Beginning in release 8.1, PostgreSQL has the ability to combine multiple indexes (including multiple uses of the same index) to handle cases that cannot be implemented by single index scans.`

priv/repo/migrations/20190308180634_soft_deletion.exs
```diff
@@ -0,0 +1,23 @@
+defmodule Companies.Repo.Migrations.SoftDeletion do
+  use Ecto.Migration
+
+  def change do
+    alter table(:companies) do
+      add(:removed_pending_change_id, references(:pending_changes))
```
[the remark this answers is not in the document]
brain-geek: yup, will change.

Well, I don't think we need anything except what pending change id. If not - let's add it later on.

brain-geek:
@doomspork I'll have comments fixed today.

brain-geek:
Hm, @doomspork , I don't think this approach differs much from one proposed by @gemantzu .

This change only adds a field to mark a record as deleted. Only data attached there is why (pending change reference). `My take on this issue would be to add an active field` would basically do the same as the current implementation.

Also, if someone deleted a company - that's for a good reason. Let them create a new one instead, IMO.

test/companies/jobs_test.exs
```diff
@@ -14,6 +14,27 @@ defmodule Companies.JobsTest do
       insert_list(2, :job)
       assert 2 == length(Jobs.all())
     end
+
+    test "does not retrieve deleted records" do
+      %{id: deleted_id} = insert(:deleted_job)
+      insert_list(2, :job)
+
+      refute deleted_id in Enum.map(Jobs.all(), & &1.id)
```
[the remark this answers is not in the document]
brain-geek: Mix format forces these to not have them. 

I suggest leaving this as-is.

test/companies/jobs_test.exs
```diff
@@ -3,7 +3,7 @@ defmodule Companies.JobsTest do
 
   import ExUnit.CaptureLog
 
-  alias Companies.Jobs
+  alias Companies.{Jobs}
```
doomspork: We don't need `{}` here since we're only aliasing a single module.

test/companies/jobs_test.exs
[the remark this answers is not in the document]
doomspork: TIL, thanks for the heads up 👍

test/companies/jobs_test.exs
brain-geek: Fixed

brain-geek:
@doomspork Fixed. Let's wait for other reviews.

lib/companies/jobs.ex
```diff
@@ -18,7 +18,10 @@ defmodule Companies.Jobs do
 
   """
   def all do
-    Repo.all(Job)
+    (j in Job)
+    |> from()
+    |> where([j], is_nil(j.removed_pending_change_id))
```
doomspork: In addition to check for non-deleted jobs, should we check for jobs belonging to non-deleted companies?

gemantzu:
I see an error when trying to manually test the functionality.
Creating a company and approving it works.
If I make an edit or a delete pending change (both companies and jobs), when I try to view it to either approve or reject, it fails with the following error:
`cannot encode association :removed_pending_change from Companies.Schema.Company to JSON because the association was not loaded.`
`cannot encode association :removed_pending_change from Companies.Schema.Job to JSON because the association was not loaded.`

Do you see them in your local repo?

doomspork:
I will test this out this weekend @brain-geek and see if I can reproduce @gemantzu's aforementioned errors 👍

brain-geek:
Sorry for slow responses. I've seen the feedback, will be fixing it this or next weekend.

doomspork:
No worries @brain-geek, take your time. Thanks again for the help!  Happy Friday 😁

doomspork:
@maartenvanvliet do you mind taking a peek at this PR? @brain-geek is getting back into open source and wants to wrap up these two outstanding out PRs of his (the other is draft #483). Thank you again for all your help :purple_heart:

doomspork:
@brain-geek I fixed the failing tests and the conflict locally. If you give me contributor access to your fork I can push the commit over to ya so we can get this merged :grin:

brain-geek:
@doomspork Wow, thanks a lot! Cherry picked and merged.

lib/companies/jobs.ex
maartenvanvliet: Yes this seems live a valid concern. If the company is soft-deleted, its jobs should *not* show up as well

lib/companies/jobs.ex
```diff
@@ -42,7 +43,12 @@ defmodule Companies.Jobs do
       ** (Ecto.NoResultsError)
 
   """
-  def get!(id), do: Repo.get!(Job, id)
+  def get!(id) do
+    (j in Job)
```
maartenvanvliet: Same as above, what if the company is soft-deleted. Its job should not be available.

lib/companies/schema/company.ex
```diff
@@ -13,7 +13,8 @@ defmodule Companies.Schema.Company do
     field :url, :string
 
     belongs_to :industry, Industry
-    has_many :jobs, Job
+    has_many :jobs, Job, defaults: [removed_pending_change_id: nil]
+    belongs_to :removed_pending_change, PendingChange
```
maartenvanvliet: Perhaps a comment on what the `removed_pending_change` does. It's not entirely clear when seeing the schema definition.

lib/companies/schema/industry.ex
```diff
@@ -2,12 +2,13 @@ defmodule Companies.Schema.Industry do
...
   import Ecto.Changeset
 
-  alias Companies.Schema.Company
+  alias Companies.Schema.{Company, PendingChange}
 
   schema "industries" do
     field :name, :string
 
     has_many :companies, Company
+    belongs_to :removed_pending_change, PendingChange
```
maartenvanvliet: Same as above regarding comment

lib/companies/schema/job.ex
```diff
@@ -2,14 +2,15 @@ defmodule Companies.Schema.Job do
...
-  alias Companies.Schema.Company
+  alias Companies.Schema.{Company, PendingChange}
 
   schema "jobs" do
     field :title, :string
     field :url, :string
     field :remote, :boolean
 
     belongs_to :company, Company
+    belongs_to :removed_pending_change, PendingChange
```
maartenvanvliet: Same as above regarding comment

doomspork:
Thank you so much for taking a look @maartenvanvliet :purple_heart: 

@brain-geek you're most welcome buddy! Always happy to help ya :grin: Let me know if you want a hand with @maartenvanvliet's comment :+1:

brain-geek:
@doomspork Thanks for help! It feels like cheating. 🔥

lib/companies/schema/job.ex
brain-geek: Done

lib/companies/schema/industry.ex
brain-geek: Done

lib/companies/schema/company.ex
brain-geek: Done

brain-geek:
@maartenvanvliet Could you look at this PR once again? I've applied your changes.

@doomspork Thanks for the help! Any ideas what also needs to be changed before merging?

Issue #469, opened 2019-03-03
Rethink edit and delete buttons


doomspork:
The buttons are currently too big and take up too much valuable real estate on each card.  How can we improve the UX while retaining the functionality?

![image](https://user-images.githubusercontent.com/73386/53703154-d3a56c80-3dcb-11e9-88a1-7c82d69170ac.png)

doomspork:
@burden / @tajchumber as the resident UX experts any thoughts?

tajchumber:
### Solution 1

Remove the text and leave the icons only, we would have to add `title` tags to the `a` elements so that some text would show on hover.

### solution 2

Move the edit and delete icons in a dropdown menu on the top right of the card see rough sketches below:
![1](https://user-images.githubusercontent.com/16062635/53703470-53eac280-3e0a-11e9-8f2a-16fd77886d7a.jpg)

![2](https://user-images.githubusercontent.com/16062635/53703471-56e5b300-3e0a-11e9-83b4-58312d586e36.jpg)

doomspork:
Solution 2 gets my vote. @gemantzu / @burden?

doomspork:
This is also an issue for jobs: 
![image](https://user-images.githubusercontent.com/73386/53704298-cb076300-3dd8-11e9-9466-94e62b4f497f.png)

doomspork:
Should the edit company page also allow you to manage jobs to avoid those individual controls?

What do you think of making cards bigger?  1 card == 2 existing cards. It would make more space for the description.

gemantzu:
I think the second solution is better, and we could take a similar approach on the jobs. Regarding making them bigger, probably a 3 per page solution would make more sense, making them twice as big might be too much.

tajchumber:
option 2 it is!

I shall open a PR for this this weekend!

PR #489, opened 2019-03-09
feat: update edit and delete buttons for company cards, #469

tajchumber:
Edit and Delte buttons for company cards are now situated in a dropdown menu in the top right section of the card

(Also contain a bit of cleaning up of the css and js files).

Although I had considered this task completed, just before I committed I realised that the dropdown menu is hidden by the "Hiring" ribbon. See screenshot below:

Screenshot 2019-03-09 at 12 33 11

What shall we do to address this? Any suggestions @doomspork @gemantzu @burden ? :)

### DIFF
--- a/assets/css/libs/_ribbon.sass
+++ b/assets/css/libs/_ribbon.sass
@@ -67,12 +67,12 @@ $ribbon-border: .1rem solid $grey-lighter !default
 
 .corner-top-right-ribbon
   top: -22px
-  right: 25px
+  left: -2px
 
 .corner-top-right-ribbon span
   top: 30px;
   left: -25px;
-  transform: rotate(45deg);
+  transform: rotate(-45deg);
 
   // Sizes
   &.is-small
--- a/assets/css/libs/_ribbon.sass
+++ b/assets/css/libs/_ribbon.sass
@@ -65,11 +65,11 @@ $ribbon-border: .1rem solid $grey-lighter !default
   text-shadow: 0 1px 1px rgba(0,0,0,.2)
   text-align: center
 
-.corner-top-right-ribbon
+.corner-top-left-ribbon
   top: -22px
   left: -2px
 
-.corner-top-right-ribbon span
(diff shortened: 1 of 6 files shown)


zacck:
Niiiiice I like this version looks cleeaaan

doomspork:
@tajchumber I'll ecto @zacck's feedback: can't we move the dropdown to the left?

tajchumber:
We can move either one, wanted you guys feedback to decide which one! 😊

gemantzu:
I think we should move the hiring on the left and stick the menu on the right. What are we going to do with the options for the jobs?

doomspork:
@tajchumber these changes are awesome. I will defer to @gemantzu and @burden for review 👍

assets/css/libs/_ribbon.sass
```diff
@@ -67,12 +67,12 @@ $ribbon-border: .1rem solid $grey-lighter !default
 
 .corner-top-right-ribbon
```
lccezinha: @tajchumber rename this css class, now the component will be placed at the left side of the div, so does not make sense anymore the name be `corner-top-RIGHT-ribbon`

assets/css/libs/_ribbon.sass
```diff
@@ -67,12 +67,12 @@ $ribbon-border: .1rem solid $grey-lighter !default
...
   top: -22px
-  right: 25px
+  left: -2px
 
 .corner-top-right-ribbon span
```
lccezinha: Same goes here.

assets/css/libs/_ribbon.sass
tajchumber: of course! thanks picking this up!

doomspork:
@tajchumber could you resolve the conflict? @gemantzu / @burden / @lccezinha any final thoughts on this?

tajchumber:
@doomspork should be all good to go! waiting for the review 👍

PR #492, opened 2019-03-09
Improve typography

Nitrino:
Just a suggestion for improving typography with a better pair of fonts.
The goal is to make pages more readable and elegant.

Before: 
localhost_4000_before

After:
localhost_4000_after


doomspork:
Thank you @Nitrino! Keep 'em coming 😁

Issue #493, opened 2019-03-10
Gettext support


doomspork:
Is it worth us considering implementing Gettext into this project? It will serve as an example for the community in addition to allowing folks to provide translations if they so choose.

Thoughts @tajchumber, @gemantzu, @Nitrino?

Nitrino:
I now think about it and planned to create an issue. 

I am the co-organizer of Moscow meetup and I often hear questions about how many and which companies in Russia write on the elixir. Therefore, I thought to launch elixir-companies.ru but found this project. 

I would like to think about how we can launch these projects together and allow local communities to be allocated to separate domains. For starters, I can add Gettext.

What do you think about this?

doomspork:
We'd had tremendous success with translations over at [elixirschool/elixirschool](https://github.com/elixirschool/elixirschool/pulls) so I know the community is excited, I think gettext is a great addition! We should definitely kick that off.

I'm not sure I understand the need for extra domains though? Wouldn't we want a single domain for the most benefit to SEO so the resource is easier to find? I'd think we could set the locale based on the user's browser setting or with path (i.e. `elixir-companies.com/ru/`). There is also the possibility of subdomains but that can hurt SEO as I understand it.

Any thoughts @brain-geek, @gemantzu, and folks?

Nitrino:
One more question:
What will we do with the description of the companies? They are filled in by users and only in one language.

doomspork:
We have #497 and the work done in #495, closing this.

Issue #252, opened 2017-12-13
Consolidating Categories


doomspork:
We have _a lot_ of categories, too many.   Should we consolidate the categories into a smaller set with broader categories?  If we do, should we add in the concept of tags?

@stevedomin any thoughts?

stevedomin:
@doomspork sorry for not responding earlier, been travelling a lot.

I guess my first question would be: what do we use categories for?

doomspork:
No worries @stevedomin!  I've been away a lot myself lately.  The categories are used for organization.  Perhaps tags would be better suited for that?

doomspork:
@burden and @ybur-yug2 either of you have any thoughts?  I know we've briefly discussed this before.

stevedomin:
@doomspork sorry my question wasn't very clear. I know where we used them but I'm more interested in why we have them in the first place? 

In the markdown we used it for grouping companies but now they're listed alphabetically and we don't offer any kind of sorting/grouping.

I guess tags would work here but I worry it would become a mess.

How do we want users to use them? What do we want to highlight with these categories/tags?

doomspork:
Ah ha!  Sorry about that @stevedomin.  I'm trying to think of a new UI for the site that would allow you to sort companies by an industry, which is essentially what categories are supposed to be.

stevedomin:
@doomspork makes sense. Should we use a list like this one then? https://gist.github.com/mbejda/19012b99a12e9d014389

Or any other standardised list of industry?

doomspork:
@stevedomin this is _awesome_, how the heck did you find that?  This would be a fantastic list to use.

stevedomin:
Not sure how I found it tbh :)

Shall I make a PR with this new list?

doomspork:
Please do @stevedomin!

stevedomin:
Love the new page @doomspork !!! <3

doomspork:
Thank you @stevedomin, @burden did most of the layout work!

PR #495, opened 2019-03-11
Add support gettext

Nitrino:
Related with #493 

### Notes:

1. Did not add the gettext to the admin area, I think we do not need to translate it. Agree?
2. The project has `resources "/users", UserController` and templates, but does not have ` UserController`. I did not add a gettex to them, it seems they need to be deleted?
3. The list of industries is stored in the database, we need to think about its localization.

Quick sample translations:
localhost_4000_ (2)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -28,7 +28,6 @@ config :companies,
   notifier: Notify.Console,
   site_data: %{
     name: "Elixir Companies",
-    description: "A collection of companies using Elixir in production.",
     maintainers: [
       "doomspork",
       "burden",
--- a/lib/companies_web/helpers/site_helpers.ex
+++ b/lib/companies_web/helpers/site_helpers.ex
@@ -1,12 +1,14 @@
 defmodule CompaniesWeb.SiteHelpers do
   @moduledoc false
 
+  import CompaniesWeb.Gettext
+
   def query_params(%{query_params: %{"type" => type}}), do: [type: type]
   def query_params(_), do: []
 
   def maintainers, do: Map.get(site_data(), :maintainers)
 
-  def site_description, do: Map.get(site_data(), :description)
+  def site_description, do: gettext("A collection of companies using Elixir in production.")
 
   def site_title, do: Map.get(site_data(), :name)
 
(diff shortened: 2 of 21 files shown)


doomspork:
Howdy @Nitrino! This is _awesome_

1. Agreed. This part of the site needs some work anyways. 👍  I'd rather no waste people's time making it better.
1. We should delete that!
1. This is a good point. We need to think about this one a bit for a couple of reasons:
    1. If anything we want to make this list smaller; consolidate some industries. We've had a free issues opened about this in the past #252 
    1. Moving the industries into a file makes it easier for folks to think they can add their own industries, which happened _all the time_ before.
    1. Lastly, I'd very much like to avoid "wasting" translators time with things subject to too much flux.

gemantzu:
@nitrino Thank you so much for your great effort. I am off decompressing a bit, I am going back to business tomorrow. I will check it up and all the other stuff that ppl are pushing towards us, I have to say, things are looking really good for this project :)

brain-geek:
@doomspork Actually, if we want to limit industries, we can just hardcode it. This way nobody will create a new ones. :trollface: 

We can take some industry list tested by time, [like this one](https://www.nasdaq.com/screening/industries.aspx). Also, this will allow translating industries more easily.

doomspork:
@brain-geek I fear that list is too small. What do you think about merging this and continuing the conversation RE: consolidating the list + moving it to a translatable place into a new issue?

brain-geek:
@doomspork That would be an awesome idea! I support that. @Nitrino did a really good job, kudos to him!

Nitrino:
Guys, I solved conflicts with master

doomspork:
@Nitrino / @brain-geek now we need to create a few more issues, correct me if I'm wrong:

1. A strategy to identify and set the locale (url path, browser headers, where in the request flow...)
1. Translations 🎉

Nitrino:
@doomspork Yep, right. 
More:
3. The ability to localize the description of companies. Need to discuss.
4. Localization of the list of industries.

doomspork:
@Nitrino I'll get those created and tag you for comments 😁 

Thanks so much for all your help

PR #502, opened 2019-03-18
Fix authorization security issue

Nitrino:
I noticed that we have authorization, but any user can edit any data without authorization and maintainer access.

Example: 
I go to the company [edit page](https://elixir-companies.com/auth/jobs/1/edit) without authorization and I have access to editing.
image

### DIFF
--- a/lib/companies_web/router.ex
+++ b/lib/companies_web/router.ex
@@ -32,13 +32,13 @@ defmodule CompaniesWeb.Router do
     get "/signout", AuthController, :signout
     get "/github", AuthController, :request
     get "/github/callback", AuthController, :callback
-    resources "/companies", CompanyController
-    resources "/jobs", JobController
   end
 
-  scope "/admin", CompaniesWeb.Admin do
-    pipe_through [:browser]
+  scope "/admin", CompaniesWeb do
+    pipe_through [:browser, :auth]
 
-    resources "/pending", PendingChangeController
+    resources "/pending", Admin.PendingChangeController
+    resources "/companies", CompanyController
+    resources "/jobs", JobController
   end
 end


doomspork:
@Nitrino this is intentional. https://github.com/doomspork/elixir-companies/blob/master/lib/companies/companies.ex#L95-L99

Create, edit, and delete of companies and jobs trigger a notification to admins to review the change and approve it.

doomspork:
We do not want to be solely responsible for maintaining the validity of this data. This allows the community contribute and only requires a quick review and approval, much like we previously used when this was a Jekyll app and required PRs for additions/updates/removals.

doomspork:
image

Nitrino:
@doomspork, Sorry, I did not notice this, but still remain two points:
1. Without authorization, the user should not see the edit page? When saving, he will fail because there is no `user_id`.
2. `https://elixir-companies.com/admin/pending` not protected too. It should be?

doomspork:
@Nitrino I noticed those while looking into something else and fixed them on master just a few minutes ago. I fixed a couple of things in the routes:

1. Moved `/companies` and `/jobs` out from under `/auth`, not sure how they got there.
2. Added `except: [:index, :show]` to both aforementioned resources, we don't have those pages
3. Moved the aforementioned resources under a `scope` with the `:auth` pipeline.
4. Added `:auth` to the admin routes 😰

Nitrino:
@doomspork You have in master. These routes do not need authorization
https://github.com/doomspork/elixir-companies/blob/master/lib/companies_web/router.ex#L29

Nitrino:
More precisely, the authorization now checks if the user has maintainer rights.
For these pages, it should check only that the current_user exists.

doomspork:
Ah ha! Good catch @Nitrino. You are correct.  See: https://github.com/doomspork/elixir-companies/pull/503

PR #503, opened 2019-03-18
Separate authenticate and authorize plugs

doomspork:
@Nitrino would you mind looking at this?  Thank you 😁

### DIFF
--- a/lib/companies_web/plugs/authorize.ex
+++ b/lib/companies_web/plugs/authorize.ex
@@ -10,7 +10,11 @@ defmodule CompaniesWeb.Plugs.Authorize do
 
   def init(opts), do: opts
 
-  def call(%{assigns: %{current_user: %{maintainer: true}}} = conn, _opts) do
+  def call(%{assigns: %{current_user: %{maintainer: true}}} = conn, maintainer: true) do
+    conn
+  end
+
+  def call(%{assigns: %{current_user: %{}}} = conn, []) do
     conn
   end
 
--- a/lib/companies_web/router.ex
+++ b/lib/companies_web/router.ex
@@ -18,6 +18,10 @@ defmodule CompaniesWeb.Router do
     plug CompaniesWeb.Plugs.Authorize
   end
 
+  pipeline :admin do
+    plug CompaniesWeb.Plugs.Authorize, maintainer: true
+  end
+
   scope "/", CompaniesWeb do
     pipe_through [:browser]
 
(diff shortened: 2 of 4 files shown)


lib/companies_web/plugs/authenticate.ex
```diff
@@ -0,0 +1,24 @@
...
+  """
+
+  import Phoenix.Controller, only: [put_flash: 3, redirect: 2]
+  import Plug.Conn
+
+  alias CompaniesWeb.Router.Helpers, as: Routes
+
+  def init(opts), do: opts
+
+  def call(%{assigns: %{current_user: %{}}} = conn, _opts) do
```
doomspork: Checks that we have a `current_user` that is a map. That's a simple enough check I think.

lib/companies_web/router.ex
```diff
@@ -15,6 +15,10 @@ defmodule CompaniesWeb.Router do
   end
 
   pipeline :auth do
+    plug CompaniesWeb.Plugs.Authenicate
```
doomspork: Instead of a new plug we could update the previous to support a `maintainer: true` option for admin pipeline.

Nitrino: I agree it will be better to update the existing plug like:
```elixir
  def call(%{assigns: %{current_user: %{}}} = conn, []) do
    conn
  end

  def call(%{assigns: %{current_user: %{maintainer: true}}} = conn, maintainer: true) do
    conn
  end
```

doomspork:
@Nitrino updated. I'm going to start adding controller tests this week. Hopefully these things are caught by tests.

lib/companies_web/plugs/authorize.ex
```diff
@@ -10,7 +10,11 @@ defmodule CompaniesWeb.Plugs.Authorize do
 
   def init(opts), do: opts
 
-  def call(%{assigns: %{current_user: %{maintainer: true}}} = conn, _opts) do
+  def call(%{assigns: %{current_user: %{maintainer: true}}} = conn, maintainer: true) do
+    conn
+  end
+
+  def call(%{assigns: %{current_user: %{}}} = conn, _opts) do
```
Nitrino: This pattern matching will allow access to the admin panel to all authorized users.
We need to check that the opts are empty.
```suggestion
  def call(%{assigns: %{current_user: %{}}} = conn, []) do
```

doomspork: Good catch!

doomspork:
@Nitrino I updated the PR and added test coverage for the Plug.

Nitrino:
@doomspork looks good now.

PR #504, opened 2019-03-19
Identify and set a user's locale for localization

### LINKED ISSUE #499: Identify and set a user's locale for localization
#493 is merged so now we need to devise a way to identify the user's locale and where to handle this. We could use URL paths (/ru/, /en/), rely on a cookie, or browser headers.

What do you think @tajchumber, @gemantzu, @Nitrino, and @brain-geek?

Nitrino:
Resolve #499

image


gemantzu:
Thanks @Nitrino !
Isn't it possible to set the locale in a Plug, so we don't have to manually insert it every time in our links? And even without using a cookie (for the sake of that damned GDPR).
Problem is, we will for sure forget about inserting that locale in a new link at some point in the future.

Nitrino:
It was worth discussing it before implementation, but I wanted to make a draft.

There are several ways to set user locale: 
1. As part of the link, or part of the path(http://elixir-companies.com/en) or part of the parameters(http://elixir-companies.com?locale=en).
2. Saving the selected locale in cookies (hi GDPR).
3. Request header `accept-language`. These are language definitions by the user system. When switching, you still need a way to save. 
4. Detect locale by IP (please do not do it)

Maybe I forgot some or I don’t know.
The current version supports all of these options with fallback:
1. Locale in URL
2. Cookie
3. Request header `accept-language`
3. Default locale from config

My goals in implementation:
1. Ability to share a link to a specific locatе (It may not be necessary)
2. Give the opportunity not to lose locale when switching between devices (iOS Handoff, just send a link via messenger to a device with another locale)


>Problem is, we will for sure forget about inserting that locale in a new link at some point in the future.

It is not possible, as the locale is indicated by the parameter to the route.
https://github.com/doomspork/elixir-companies/blob/92f4b1f4cc10b2132bdbbdb6f6546bf13b4f552e/lib/companies_web/router.ex#L33
And if you forget to pass it on, the Phoenix will return an error and show URL helpers signatures.
The only option where this is possible is to use links without using Phoenix helpers.

Maybe I'm missing something and there’s a better way to implement it.

gemantzu:
I was thinking I could somehow intercept a path method but I have to check if it is possible at all. But since it will throw an error if someone forgets about it, we might be ok as is. @doomspork thoughts?

gemantzu:
I don't want this to freeze, I can retest it this afternoon when I am back home and we can proceed with the merge. If we are able to, we can improve it in the future as always.

lib/companies_web/router.ex
```diff
@@ -35,19 +43,17 @@ defmodule CompaniesWeb.Router do
       resources "/companies", CompanyController, except: [:index, :show]
       resources "/jobs", JobController, except: [:index, :show]
     end
+
+    scope "/admin", Admin do
```
doomspork: I don't know that this will ever be translated, does it need to live under the locale?

lib/companies_web/controllers/admin/pending_change_controller.ex
```diff
@@ -20,14 +20,14 @@ defmodule CompaniesWeb.Admin.PendingChangeController do
       {:ok, _approved_changes} ->
         conn
         |> put_flash(:info, "Changes submitted")
-        |> redirect(to: Routes.pending_change_path(conn, :index))
+        |> redirect(to: Routes.pending_change_path(conn, :index, conn.params["locale"]))
```
doomspork: If possible I would like to try and refactor out the `conn.params["locale"]` calls that come with this PR. I think there are two ways we can accomplish this:

1. Pattern matching in the function head

```elixir
def update(conn, %{"id" => change_id, "approval" => approval, "locale" => locale}) do
  case PendingChanges.approve(change_id, approval == "true") do
    {:ok, _approved_changes} ->
      conn
      |> put_flash(:info, "Changes submitted")
      |> redirect(to: Routes.pending_change_path(conn, :index, locale))

    {:error, reason} ->
      Logger.error(reason)

      conn
      |> put_flash(:error, "Changes failed")
      |> redirect(to: Routes.pending_change_path(conn, :index, locale))
  end
end
```

2. With a helper function (`UserHelpers` perhaps?):

```elixir
def update(conn, %{"id" => change_id, "approval" => approval}) do
  case PendingChanges.approve(change_id, approval == "true") do
    {:ok, _approved_changes} ->
      conn
      |> put_flash(:info, "Changes submitted")
      |> redirect(to: Routes.pending_change_path(conn, :index, locale(conn)))

    {:error, reason} ->
      Logger.error(reason)

      conn
      |> put_flash(:error, "Changes failed")
      |> redirect(to: Routes.pending_change_path(conn, :index, locale(conn)))
  end
end


defmodule CompaniesWeb.UserHelpers do
  def locale(%{params: %{"locale" => locale}}), do: locale
  def locale(_), do: "en"
end
```

A helper is probably preferably as we can use it in our views to remove the same `@conn.params["locale"]` code there.

doomspork: We could also access the `:locale` assign in place of `params`, directly, which would rely on `SetLocale` as we do elsewhere: https://github.com/smeevil/set_locale/blob/master/lib/set_locale.ex#L37

mix.exs
```diff
@@ -53,7 +53,8 @@ defmodule Companies.MixProject do
       {:bypass, "~> 1.0", only: :test},
       {:ex_machina, "~> 2.2", only: :test},
       {:excoveralls, "~> 0.10", only: :test},
-      {:phoenix_live_reload, "~> 1.2", only: :dev}
+      {:phoenix_live_reload, "~> 1.2", only: :dev},
+      {:set_locale, "~> 0.2.1"}
```
doomspork: Could you move this above the `only:` section? We try to follow similar community standards with test/dev only dependencies at the bottom of the list and the rest above those, each group sorted alphabetically 😁

lib/companies_web/router.ex
```diff
@@ -25,6 +26,13 @@ defmodule CompaniesWeb.Router do
   scope "/", CompaniesWeb do
     pipe_through [:browser]
 
+    get "/", CompanyController, :recent
```
doomspork: Will we ever hit these routes with the `SetLocale` plug?

lib/companies_web/router.ex
Nitrino: We have navigation links to translated pages in the admin panel. For their work, we need to pass the current locale. For such cases, it is easier to that all pages are under the locale.

lib/companies_web/router.ex
Nitrino: I use this solution.

lib/companies_web/controllers/admin/pending_change_controller.ex
Nitrino: I use this solution.

lib/companies_web/router.ex
Nitrino: In the current implementation using locale helper and the default locale, we can move the admin panel from under the locale, but the locale will be lost when going through the admin panel.
I think it's better to leave it under the locale.

doomspork:
> Sorry for the delay with this PR, I needed to mental rest.

No worries 😁 No deadlines here so please do take your time when you need it.

> I added a helper to determine localization.
> I don’t really like what this solution looks like, it seems there should be a way to make it easier, but I don’t see it yet.
> If you have ideas, I’ll be happy to hear them or we can improve this logic in the future.

Agreed! I think this cleaned things up a bit, we can figure out a cleaner solution later. For now this looks good to me.

@brain-geek / @gemantzu do you want to take one last peek at this before we merge?

Nitrino:
Updated to the current master.

PR #507, opened 2019-03-29
Move distinct/2 into individual predicate functions

doomspork:
I noticed after @tajchumber's most recent change that the homepage was being sorted alphabetically when they should be sorted by the inserted_at date.  Upon further inspection I noticed we had `order_by/3` already but the `distinct/2` at the top was throwing things off. This PR should resolve that.


gemantzu:
:+1: :+1: :+1:

PR #508, opened 2019-03-31
scrollbar brought back for overflowing cards

gemantzu:
At some point we lost the ability to scroll company cards which overflow the cards. This fixes it.
@tajchumber / @burden thoughts on improving this? maybe style a bit the scrollbar?

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -30,6 +30,7 @@ html {
   background-attachment: fixed, fixed;
   background-image: asset_url("overlay.png"), asset_url("paper_fibers.png");
 
+
   body {
     display: flex;
     min-height: 100vh;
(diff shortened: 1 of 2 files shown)


gemantzu:
I just copied and pasted the gist, because Idk enough to tinker with it. I can either remove it and let you do the proper work, or you can do your thing after this is merged :)

doomspork:
@tajchumber and @burden, can you take another look?

assets/css/app.scss
```diff
@@ -30,6 +30,7 @@ html {
   background-attachment: fixed, fixed;
   background-image: asset_url("overlay.png"), asset_url("paper_fibers.png");
 
+
```
tajchumber: All looks good to me! Could remove the extra line here please?

assets/css/app.scss
```diff
@@ -98,10 +99,36 @@ html {
   }
 
   .box {
+
+
```
tajchumber: en here as well

tajchumber:
thanks @gemantzu !

Issue #510, opened 2019-04-03
Add Resultados Digitais as a company on the list


lccezinha:
I would like to add Resultados Digitais as a company on the list.

Our main stack is Ruby and Go, but we have some projects that are built with Elixir/Phoenix.

Company data:

```
name: Resultados Digitais
description: We develop a Marketing Automation Platform for S&M companies
url: https://resultadosdigitais.com.br/
github: https://github.com/ResultadosDigitais/
localhost: Florianópolis, Brasil
blog: http://shipit.resultadosdigitais.com.br/
industry_id: Enterprise SaaS
``` 

Is that enough? Also, I was not able to find a way to contribute to and add new companies.

gemantzu:
Hi @lccezinha ! As you can see in the README, this repo has changed the past couple of months. We have moved away from the previous implementation (Ruby - Jekyll) where we had to do a PR to enlist a company, to a Phoenix based full blown database-backed application. You can visit the [website](https://elixir-companies.com), login in with your Github account and add a company through the form. We will verify your data and publish what you have entered usually within the day.

You can still contribute to the project if you want to, as it has remained 100% open source.

lccezinha:
@gemantzu Thanks, I already sent the information about the company, thank you.

lccezinha:
Also, maybe this information could be more explicit in the readme file, I did not find it in any place.

doomspork:
@lccezinha would you care to open a PR updating the README? 😁

lccezinha:
@doomspork I'll do that! Thank you!

doomspork:
Thank _you_ @lccezinha! It's awesome to have others getting involved. Have a great weekend 🎉

PR #511, opened 2019-04-05
Add instructions to add new companies

lccezinha:
As a result about the discussion with @doomspork here https://github.com/doomspork/elixir-companies/issues/510#issuecomment-480352693 I improve the read me with instructions to add new companies.


README.md
```diff
@@ -8,7 +8,17 @@ A [collection of companies using Elixir](https://elixir-companies.com/) in produ
 
 Proudly built with [Phoenix](https://phoenixframework.org).
 
-## Running
+### Adding a new company to the list
```
gemantzu: We can also add jobs on companies, can you mention that? Thank you :)

lccezinha: @gemantzu Done!

PR #513, opened 2019-04-14
Email notifications

doomspork:
@zacck apologizes for the delays but I've finally had time to finish the email setup for SendGrid. I pulled your branch and made a few changes to get Sendgrid emails working:

image

Would you mind taking a final look at this before we merge?

@gemantzu @tajchumber any thoughts?


zacck:
@doomspork  looks good! kudos for going the extra mile to add test coverage for the work!

mix.exs
```diff
@@ -36,7 +36,7 @@ defmodule Companies.MixProject do
   defp deps do
     [
       {:appsignal, "~> 1.0"},
-      {:bamboo, "~>1.2.0"},
+      {:bamboo, github: "thoughtbot/bamboo"},
```
zacck: aaaah so the one on master supports dynamic templates!

lib/companies_web/router.ex
```diff
@@ -29,6 +29,10 @@ defmodule CompaniesWeb.Router do
   scope "/", CompaniesWeb do
     pipe_through [:browser, :set_locale]
 
+    if Mix.env() == :dev do
```
notactuallypagemcconnell: Not in love with dev-specific code that is in the main code. But its whatever. Curious question: Will the compiler ditch this whole block when compiling with MIX_ENV=prod since it would evaluate to false?

lib/notify/email.ex
```diff
@@ -0,0 +1,49 @@
...
+  """
+  import Bamboo.Email
+  import Bamboo.SendGridHelper
+
+  @doc """
+  This function builds an email that informs a user of their attempted changes being approved and applied
+  """
+  @spec change_approved(map()) :: Bamboo.Email.t()
+  def change_approved(pending_change) do
+    template_id = System.get_env("SENDGRID_APPROVED_TEMPLATE_ID")
```
notactuallypagemcconnell: I believe we should prefer `Application.get_env/2` here

lib/companies_web/router.ex
doomspork: Me either. I'm otherwise sure how to handle this one specific dev route.

I do not believe the compiler will remove this call.

lib/notify/email.ex
doomspork: Why? This value will not be checked into source control because it's a unique value to our Sendgrid account.

If we put in `Application` we'll loose the ability to change the ENV on-the-fly.

lib/companies_web/router.ex
doomspork: Any suggestions?

mix.exs
doomspork: This has now been released as `1.3`

doomspork:
@gemantzu / @tajchumber could you review the Elixir? The functionality has been tested locally and works 👍

lib/companies_web/plugs/redirect.ex
```diff
@@ -1,4 +1,4 @@
-defmodule CompaniesWeb.Redirect do
+defmodule CompaniesWeb.Plugs.Redirect do
```
doomspork: This plug may not be necessary.

lib/companies/email_events.ex
```diff
@@ -0,0 +1,53 @@
...
+        case update_user_preferences(event, user) do
+          {:ok, updated_user} -> updated_user
+          _ -> user
+        end
+
+      Map.put(acc, user_email, updated_user)
+    else
+      acc
+    end
+  end
```
gemantzu: Maybe turn this into a with statement to reduce complexity?

doomspork: Done :+1:

lib/companies/email_events.ex
```diff
@@ -0,0 +1,57 @@
...
+      false ->
+        :ok
+
+      {:error, %{errors: errors}} ->
+        send_error(errors, params)
+    end
+  end
+
+  defp send_error(errors, params) do
+    Appsignal.send_error(%RuntimeError{}, "SendGrid Callback", [], %{params: params, errors: errors})
```
doomspork: If something does go sideways in our asynchronous write, let's tell AppSignal about it.

doomspork:
@gemantzu give this one more thoroughly look through (and local test if you so desire) and let's get this merged.

Part of this PR adds the user preference for receiving notifications via email. We could add a user account page to allow them to toggle this. An account section could have all kinds of fun features like a list of all the user's submitted changes and whether they were approved or rejected.

lib/companies/email_events.ex
gemantzu: Nice!

PR #525, opened 2019-06-08
include all names; order by name

RichMorin:
This is untested code, but I think it's pretty close... It purports to handle the first two parts of [Issue 523](https://github.com/beam-community/elixir-companies/issues/523).

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -126,8 +126,11 @@ defmodule Companies.Companies do
     Company.changeset(company, %{})
   end
 
-  defp predicates(query, %{"type" => <<start, finish>>}) when start in 97..122 do
+  defp predicates(query, %{"type" => <<start, finish>>}) do
     query = distinct(query, [c, _i, _j], asc: fragment("LOWER(?)", c.name))
+
+    start  = if (start  ==  97) do  32 else  97 end
+    finish = if (finish == 122) do 126 else 122 end
 
     Enum.reduce(start..finish, query, fn char, acc ->
       or_where(acc, [c], ilike(c.name, ^"#{[char]}%"))
@@ -139,12 +142,12 @@ defmodule Companies.Companies do
     |> exclude(:left_join)
     |> distinct([c, _i, _j], true)
     |> join(:inner, [c, _i, _j], j in assoc(c, :jobs))
-    |> order_by([_c, _i, j], desc: j.inserted_at)
+    |> order_by([_c, _i, j], asc: c.name)
   end
 
   defp predicates(query, _) do
     query
     |> distinct([c, _i, _j], true)
-    |> order_by([c, _i, _j], desc: c.inserted_at)
+    |> order_by([c, _i, _j], asc: c.name)
   end
 end


doomspork:
@RichMorin looks like the build isn't passing now. Could you please fix that up and add test coverage? 

We can't accept PRs without coverage.

lib/companies/companies.ex
```diff
@@ -126,9 +126,12 @@ defmodule Companies.Companies do
     Company.changeset(company, %{})
   end
 
-  defp predicates(query, %{"type" => <<start, finish>>}) when start in 97..122 do
+  defp predicates(query, %{"type" => <<start, finish>>}) do
     query = distinct(query, [c, _i, _j], asc: fragment("LOWER(?)", c.name))
 
+    start  = if (start  ==  97) do  32 else  97 end
```
doomspork: @RichMorin can you explain what you're trying to accomplish here? I think there may be a better way but it's not entirely clear the goal.

lib/companies/companies.ex
```diff
@@ -139,12 +142,12 @@ defmodule Companies.Companies do
     |> exclude(:left_join)
     |> distinct([c, _i, _j], true)
     |> join(:inner, [c, _i, _j], j in assoc(c, :jobs))
-    |> order_by([_c, _i, j], desc: j.inserted_at)
+    |> order_by([_c, _i, j], asc: c.name)
```
doomspork: Why did you change this? We want the newest jobs first, we don't care about the company name.

lib/companies/companies.ex
RichMorin: The `start` and `finish` munging are supposed to extend the range of items displayed. The lowest range (currently `ae`) should display companies whose names begin with any characters that sort before `a` (e.g., `0`). The highest range (currently `uz`) should display companies whose names begin with any characters that sort after `z` (e.g., `{`).

lib/companies/companies.ex
RichMorin: Point taken, though some users might want an option. In any case, the sort key should be obvious to the user. So, I'd suggest that you add the submission date to the metadata. That way, the user can easily figure out if a job has already been skimmed or is simply too old to be of interest.

lib/companies/companies.ex
doomspork: If we need to handle numbers and special characters, we should add that filter instead of putting numbers into A-E and symbols into U-Z.

RichMorin: There are companies in the data set whose names begin with digits (e.g., "500 Startups") and special characters (e.g., "{sauce}"), so I don't think there is much question about whether they need to be displayed. Where to put them is a judgement call. If you're listing the entire data set, do you put them all at the end, split them up according to the ASCII sort sequence, or what?  And, if you're displaying range-based subsets such as "A-E", do you define a range for the digits and special characters.  Your call...

doomspork:
There has been no activity on this in some time and the project has since had a number of UX updates. @RichMorin if you're still interested in making some changes, feel free to re-open this PR with the build passing :+1:

PR #526, opened 2019-08-17
WIP: Liveview

gemantzu:
This is why I hate frontend... Live loading works, but what now needs to be done is to adapt the search function to preload jobs / industries as `all/1` does.


gemantzu:
@doomspork take a look, liveview works except hiring. Plus we could use @burden 's or @tajchumber 's help on the form looks.

lib/companies/jobs.ex
```diff
@@ -102,4 +102,20 @@ defmodule Companies.Jobs do
   def change(%Job{} = job) do
     Job.changeset(job, %{})
   end
+
+  def search(filters) do
+    filters
+    |> Enum.reduce(Job, fn
+      {_, nil}, query ->
```
doomspork: Let's make this a named function instead. I've never been a big fan of anonymous functions w/ many heads.

How about something like this? I _think_ we can also reduce this to just 2 matches:

```elixir
defp query_predicates({"text", text}, query), 
  do: from c in query, where: ilike(c.title, ^"%#{text}%")
defp query_predicates(_, query), 
  do: query
```

gemantzu: Sure, I will do the change. Still adding stuff btw.

gemantzu:
I am at my limit with JS here, can't really tell why it runs like that. Thoughts @doomspork  / @tajchumber / @SophieDeBenedetto ?

gemantzu:
Closed in favor of #534 (non LiveView)

PR #527, opened 2019-09-27
Add remote option to job 

maartenvanvliet:
* Adds the remote field to the jobs table
* Shows remote option for jobs in company card
* Adds remote option to create/update form

### DIFF
--- a/lib/companies/schema/job.ex
+++ b/lib/companies/schema/job.ex
@@ -7,6 +7,7 @@ defmodule Companies.Schema.Job do
   schema "jobs" do
     field :title, :string
     field :url, :string
+    field :remote, :boolean
 
     belongs_to :company, Company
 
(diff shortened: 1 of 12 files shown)


lib/companies/schema/job.ex
```diff
@@ -7,6 +7,7 @@ defmodule Companies.Schema.Job do
   schema "jobs" do
     field :title, :string
     field :url, :string
+    field :remote, :boolean
```
pdgonzalez872: Nit: What about when a job is not just remote or not? What if it is "remote for the right candidate"?

doomspork: That's a good question @pdgonzalez872! I think a note could go in the job description. On other job sites I've seen those jobs listed as on-site with a note about remote for the right person.

Thoughts @maartenvanvliet? @gemantzu?

maartenvanvliet: don't think it needs to be covered right now, as said a note could go in a description.

gemantzu: @doomspork I think you are right, we can not cover all cases like that, but as an afterthought we could turn it into a dropdown selection ("remote", "on-site", "remote for the right person") as a new Schema (JobLocation), and we could keep adding to it as needed. 

But I believe this is not actually needed, simply adding a remote field should suffice for 99% of the use cases.

doomspork:
@maartenvanvliet thank you so much! There are plenty of improvements that could be made to the site, we'd more than welcome your assistance 😁

maartenvanvliet:
👍

PR #529, opened 2019-09-27
Add link to github repository

maartenvanvliet:
To encourage contributions to the repo I think we should link directly from the site to the repository.


doomspork:
@gemantzu let's start here and if need be, do a follow-up PR to expand on this.

doomspork:
@gemantzu let's start here and if need be, do a follow-up PR to expand on this.

gemantzu:
Thank you for your contribution @maartenvanvliet !

PR #534, opened 2019-10-05
Adds Filters for Companies / Jobs - Seperates Jobs into new route

gemantzu:


lib/companies.ex
```diff
@@ -1,9 +1,146 @@
...
+  Returns the list of paginated companies.
+
+  ## Examples
+
+  iex> all()
+  [%Company{}, ...]
+
+  """
+
+  def all(params \\ %{}) do
```
doomspork: Shouldn't all of this technically live in `Companies.Companies`?

lib/companies_web/router.ex
```diff
@@ -37,8 +37,9 @@ defmodule CompaniesWeb.Router do
     pipe_through [:browser, :set_locale]
 
     get "/", CompanyController, :recent
-    get "/hiring", Redirect, to: "/browse?type=hiring"
-    get "/browse", CompanyController, :index
+    get "/hiring", Redirect, to: "/jobs"
```
doomspork: Do we still need this? I can see the case for either.

lib/companies_web/views/company_view.ex
```diff
@@ -28,4 +28,15 @@ defmodule CompaniesWeb.CompanyView do
   def select_industries(industries) do
     Enum.map(industries, fn %{id: id, name: name} -> {name, id} end)
   end
+
+  def selected_text(%{}, _), do: ""
+  def selected_text(%{search: %{}}, _), do: ""
+
+  def selected?(params = %{search: %{"industry_id" => iid}}, industry_id) do
+    IO.inspect(params)
```
doomspork: We don't want this 👍

lib/companies_web/views/company_view.ex
```diff
@@ -28,4 +28,15 @@ defmodule CompaniesWeb.CompanyView do
   def select_industries(industries) do
     Enum.map(industries, fn %{id: id, name: name} -> {name, id} end)
   end
+
+  def selected_text(%{}, _), do: ""
+  def selected_text(%{search: %{}}, _), do: ""
```
doomspork: Do we need these if they both return `""`? Could we just put `""` where you call these?

If for some reason we do need this function, we could simplify to one I believe:
```elixir
def selected_text(_, _), do: ""
```

lib/companies_web/views/job_view.ex
```diff
@@ -17,4 +19,6 @@ defmodule CompaniesWeb.JobView do
       "input"
     end
   end
+
+  def remote_ribbon(job), do: if(job.remote, do: "has-ribbon", else: "")
```
doomspork: Can you make this multi-line so we don't need the `()` or use pattern matching?

```elixir
def remote_ribbon(%{remote: true}), do: "has-ribbon"
def remote_ribbon(_), do: ""
```

lib/companies_web/views/job_view.ex
```diff
@@ -17,4 +19,18 @@ defmodule CompaniesWeb.JobView do
...
+  def remote_ribbon(_), do: ""
+
+  def link_title(url) do
+    url
+    |> String.replace("http://", "")
+    |> String.replace("https://", "")
+    |> String.replace("www.", "")
+  end
+
+  def url_with_scheme("http://" <> _ = url), do: url
```
doomspork: Do we want to focus https for all urls?

lib/companies_web/views/job_view.ex
```diff
@@ -17,4 +19,18 @@ defmodule CompaniesWeb.JobView do
       "input"
     end
   end
+
+  def remote_ribbon(%{remote: true}), do: "has-ribbon"
+  def remote_ribbon(_), do: ""
+
+  def link_title(url) do
+    url
+    |> String.replace("http://", "")
```
doomspork: I think we can find a way to do this ideally without three separate operations. Could we use a regex?

lib/companies_web/views/company_view.ex
```diff
@@ -28,4 +28,10 @@ defmodule CompaniesWeb.CompanyView do
   def select_industries(industries) do
     Enum.map(industries, fn %{id: id, name: name} -> {name, id} end)
   end
+
+  def selected?(%{search: %{"industry_id" => iid}}, industry_id) do
```
doomspork: Can we simplify this into two function heads?


`def selected?(%{search: %{industry_id: industry_id}, industry_id)`

michaelstalker: Is this function is actually being used?

Functions ending in a question mark also return boolean values, by convention. See https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo.

gemantzu: @michaelstalker You are right, I used this function when I tried to repopulate the form with the filters already selected, but then since I couldn't make it work, I removed the use of it.

lib/companies_web/views/job_view.ex
gemantzu: We are not, we just make it the default when the user is not giving us any. (I don't remember why we originally did that for the companies though)

lib/companies_web/views/company_view.ex
doomspork: Hooray for @michaelstalker 🎉 

Hope you're well good buddy 😁

PR #537, opened 2019-10-09
Markdown format for company description

maartenvanvliet:
Fixes https://github.com/beam-community/elixir-companies/issues/536

I've added Earmark as a dependency to convert the markdown to html, furthermore HtmlSanitizeEx is used to strip out anything that is not markdown html for security purposes.


doomspork:
@maartenvanvliet could you rebase on master and add a test for the helper?

gemantzu:
Works perfectly fine just tested it locally on my machine. Add a test for the helper and we can merge!

PS: would it make sense to also note people on the create company form that markdown is available?

gemantzu:
Awesome @maartenvanvliet ! Thanks!

PR #539, opened 2019-10-10
Add company page

maartenvanvliet:
This PR adds a company page as mentioned in https://github.com/beam-community/elixir-companies/issues/533

It is very basic for now, just a page-wide version of the company card.

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -76,7 +76,13 @@ defmodule Companies.Companies do
   ** (Ecto.NoResultsError)
 
   """
-  def get!(id), do: Repo.get!(Company, id)
+  def get!(id, opts \\ []) do
+    preloads = Keyword.get(opts, :preloads, [])
+
+    Company
+    |> preload(^preloads)
+    |> Repo.get!(id)
+  end
 
   @doc """
   Submits a new company for approval.
--- a/lib/companies_web/controllers/company_controller.ex
+++ b/lib/companies_web/controllers/company_controller.ex
@@ -38,7 +38,7 @@ defmodule CompaniesWeb.CompanyController do
   end
 
   def show(conn, %{"id" => id}) do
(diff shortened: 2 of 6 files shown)


lib/companies/companies.ex
```diff
@@ -76,7 +76,13 @@ defmodule Companies.Companies do
   ** (Ecto.NoResultsError)
 
   """
-  def get!(id), do: Repo.get!(Company, id)
+  def get!(id, opts \\ []) do
```
gemantzu: TIL, thank you!

gemantzu:
I don't really like how the show page looks. any thoughts @doomspork ?

lib/companies/companies.ex
doomspork: Nice! @maartenvanvliet great minds think alike! I just implemented this pattern in another project.

It’s also possible, but more work, to work this into a query builder such that you don’t need _N+1_ queries. 

Really like this 👌

maartenvanvliet:
Screenshot 2019-10-28 at 11 25 24

In this PR it looks like this, so the card that was already used but then wider. It could be made a little smaller on desktop, and perhaps not use the card. Any other suggestions?

gemantzu:
I think we should consider moving away from the cards, at least for the company show page (maybe for the index as well, idk).

maartenvanvliet:
Screenshot 2019-10-29 at 15 56 22

This is what I came up with by removing the card :)

gemantzu:
That's already much better. @doomspork ?

doomspork:
Sorry @maartenvanvliet I have been very busy the last few weeks. I think this looks great. Could you rebase your PR so we can merge?

maartenvanvliet:
Done :)

gemantzu:
Thank you @maartenvanvliet !!!!

PR #540, opened 2019-10-10
Improve diff of edits

maartenvanvliet:
This PR improves the edit diff view. See also https://github.com/beam-community/elixir-companies/issues/533

It adds a column in the pending changes overview to see when the changes were made.

It also improves the diff view, by pretty printing the JSON and outputting them in a `<pre>` tag. Furthermore it shows additions/removals by setting the backgroundColor of the edits much like how github does it. It's still fairly basic but I think it works better
Screenshot 2019-10-10 at 20 33 48

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -71,7 +71,7 @@ document.addEventListener("DOMContentLoaded", function() {
     }
   };
 
-  var $diff = document.querySelector(".box .diff");
+  var $diff = document.querySelector(".box .diff pre");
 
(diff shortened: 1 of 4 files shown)


gemantzu:
I think this is towards the right direction. I can really tell now what most of the diffs are. Thoughts @doomspork ?

maartenvanvliet:
Sure, would love to help out!

doomspork:
Awesome!

Issue #554, opened 2019-12-09
Order by created_at / inserted_at for recent


gemantzu:
@doomspork / @maartenvanvliet which one do you think?

PR #555, opened 2019-12-09
Bugfix/recent companies sort

gemantzu:
solves #554

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -28,8 +28,13 @@ defmodule Companies.Companies do
     |> Repo.paginate(page: page)
   end
 
+  @doc """
+  Simply calls `all/1` with a two level order, so that even in the slight chance we have
+  two companies with the exact same time of insertion, they will be ordered with
+  id next.
+  """
   def recent do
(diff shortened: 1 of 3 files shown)


lib/companies/companies.ex
```diff
@@ -34,8 +34,7 @@ defmodule Companies.Companies do
   id next.
   """
   def recent do
-    %{order: [desc: :inserted_at, desc: :id]}
-    |> all()
+    all(%{order: [desc: :inserted_at, desc: :id]})
```
maartenvanvliet: good call on removing the pipe when piping with a single function. 

Would it make sense to add Credo to this project to enforce such rules across the codebase.

gemantzu: I was instructed  to do so by our beloved @doomspork  :)

doomspork: Hahaha! You're too funny @gemantzu 

@maartenvanvliet I'm open to using Credo. I find it very helpful. I'll leave it up to someone else to add to the project. I believe we use it on some other projects so we may be able to borrow the config.

PR #559, opened 2019-12-09
Gettext

gemantzu:
update + greek

### DIFF
--- a/priv/gettext/default.pot
+++ b/priv/gettext/default.pot
@@ -17,42 +17,25 @@ msgstr ""
 
 #, elixir-format
 #: lib/companies_web/templates/company/card.html.eex:58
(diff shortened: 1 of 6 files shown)


priv/gettext/default.pot
```diff
@@ -227,21 +188,107 @@ msgid "Name"
...
+msgstr ""
+
+#, elixir-format
+#: lib/companies_web/templates/job/index.html.eex:35
+msgid "Remote"
+msgstr ""
+
+#, elixir-format
+#: lib/companies_web/templates/job/show.html.eex:16
+msgid "Remote:"
```
doomspork: @gemantzu can you find where we use this and change it so it's just `Remote`, that way we don't have 2 nearly identical translations

PR #563, opened 2019-12-10
Trim whitespace from search text before query

aaronrenner:
I was searching for DockYard this morning but the iOS autocomplete appended a space creating a search like "DockYard ", causing no results to be returned. This PR trims the leading and trailing whitespace before building the database query.


gemantzu:
@aaronrenner you could also do the same for jobs if you want. Otherwise I will fix it this afternoon :)

aaronrenner:
@gemantzu Thanks for the idea! I just submitted a PR for it.

PR #566, opened 2019-12-11
Add reviewed by to pending changes

maartenvanvliet:
In the list of approved/rejected changes I missed by whom it was reviewed. This PR adds that and shows it when clicking on a reviewed change.

All previous changes are without a `reviewer_id` so nothing is shown it that situation. Not sure if backfilling is necessary or desirable.


lib/companies/schema/pending_change.ex
```diff
@@ -15,17 +15,25 @@ defmodule Companies.Schema.PendingChange do
...
+  def changeset(pending_change, attrs) do
+    pending_change
+    |> cast(attrs, [:action, :changes, :resource, :user_id])
     |> validate_required([:action, :changes, :resource, :user_id])
     |> validate_inclusion(:action, ["create", "update", "delete"])
     |> validate_inclusion(:resource, ["company", "industry", "job"])
     |> assoc_constraint(:user)
   end
+
+  def approve_changeset(pending_change, attrs) do
```
maartenvanvliet: I split up the changeset functions to one for creating the pending change, and another for approving it.

lib/companies/pending_changes.ex
```diff
@@ -175,9 +175,9 @@ defmodule Companies.PendingChanges do
   defp struct_to_string(%Industry{}), do: "industry"
   defp struct_to_string(%Job{}), do: "job"
 
-  defp update_approval(change, note, approval) do
+  defp update_approval(change, note, %{id: reviewer_id} = _reviewer, approval) do
```
gemantzu: Why do you also pattern match on an discarded var here? For readability?

maartenvanvliet: Yes, to make clear that the reviewer is passed in there.

gemantzu: Ok thanks :)

doomspork:
@maartenvanvliet you're awesome! Thank you so much for all of these improvements

PR #567, opened 2019-12-11
Add credo to project

### LINKED ISSUE #558: Add Credo to project
See https://github.com/beam-community/elixir-companies/pull/555#discussion_r355585459

I propose we add Credo to the project and CI to enforce code guidelines. 
We'll need to investigate which rules we use. The default config is fine imo, though it may be necessary to disable some rules for now. This to prevent a very noisy commit when we initially run Credo and fix its warnings.

maartenvanvliet:
To get more consistency and readability in the code across the project I've proposed in this issue (https://github.com/beam-community/elixir-companies/issues/558) to add Credo.

This PR fixes #558 
I've fixed the issues that a run of `mix credo` revealed. They were pretty minor, mostly doc fixes. Credo has also been added to CI so it will run on each push.


gemantzu:
Nice! Thank you @maartenvanvliet !!! Let's wait for @doomspork to wake up and do a double check on it, ok?

doomspork:
Awesome :boom:

PR #570, opened 2019-12-12
adds reviewer on pending_change index

gemantzu:
Should we add tests for pending_view_test module?


gemantzu:
:thinking: I don't think so. This just shows who reviewed this pending change. I probably have to change this so that it won't show up if the pending change is, well, pending

doomspork:
Ah ha! Nevermind. I see how we updated these pages.

PR #574, opened 2019-12-22
Revert "Fix search pagination in companies"

doomspork:
Reverts beam-community/elixir-companies#572

@brain-geek this PR has broken the site so I am reverting for now. We need to add some additional test to the project because this should have been caught. Right now the index page for companies is a 500ing.


brain-geek:
@doomspork Hm, that's weird - seems like I manually tested this, but not after final changes. Will be adding tests and fixing it.

doomspork:
Thanks @brain-geek! That's also on me for not testing. I need to see what I can do with Heroku to setup a staging ENV or something for us. Pulling down every PR and running it is not something I have time to do :cry:

PR #579, opened 2020-02-12
adds preload to resource pending change view

### LINKED ISSUE #578: Cannot view resource deletion
This is the error shown:
`cannot encode association :removed_pending_change from Companies.Schema.Company to JSON because the association was not loaded.`

gemantzu:
fixes #578


gemantzu:
I noticed the error and just run it locally to see what's up in detail. It is also reported on AppSignal (I just checked it, because for some reason the app has stopped sending me emails upon error).

PR #587, opened 2020-02-13
user profiles page

gemantzu:


lib/companies_web/templates/user/card.html.eex
```diff
@@ -0,0 +1,16 @@
+<div class="company box %>">
+  <div class="content has-text-centered">
+    <p class="title is-4 thin"><%= link to: Routes.user_path(@conn, :show, locale(@conn), @user) do %><%= Map.get(@user, :name) %><% end %></p>
```
maartenvanvliet: This route ` Routes.user_path(@conn, :show, locale(@conn), @user)` is not implemented right?

Do we want to expose a public route for users who are currently in the database? Or only for those who have said they are looking for a job.

lib/companies_web/controllers/auth_controller.ex
```diff
@@ -6,6 +6,7 @@ defmodule CompaniesWeb.AuthController do
   plug Ueberauth
 
   def callback(%{assigns: %{ueberauth_auth: auth}} = conn, _params) do
+    IO.inspect auth
```
maartenvanvliet:
```suggestion
```

IO.inspect call can be removed

priv/repo/migrations/20200213165634_adds_profile_fields_to_users.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Companies.Repo.Migrations.AddsProfileFieldsToUsers do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :name, :string
+      add :bio, :text
+      add :interests, :text
+      add :location, :string
```
maartenvanvliet: Perhaps we can add a :country field to the table and show a dropdown for countries on the profile form? Can be done in a followup PR though

lib/companies_web/templates/user/card.html.eex
gemantzu: Only for those who are looking for a job. This is still WIP, lot's of work to be done.

gemantzu:
@maartenvanvliet I upped this just for backup reasons (and to show you guys a glimpse of where this is going, just in case I am totally off). Still early, no tests yet for additional code, access to users must be limited to only those who are looking for a job etc.

gemantzu:
I only need to add some tests and I am probably done with this feature. It's looking good so far, please check it out and let me know if I am on the right direction.

priv/repo/migrations/20200213165634_adds_profile_fields_to_users.exs
gemantzu: We have the location field from Github, we can probably work with that but we will see how it goes.

gemantzu:
I think I am done.
@doomspork / @maartenvanvliet please check.

doomspork:
We need to expire profiles (and jobs) after a period of time :thinking:

PR #591, opened 2020-02-22
fixes profile button misplacement

gemantzu:


lib/companies/accounts.ex
```diff
@@ -16,6 +16,13 @@ defmodule Companies.Accounts do
     |> Repo.paginate(page: page)
   end
 
+  def get_developer_profile(id) do
```
doomspork: I think you could simplify this function with `get_by/3`:

```elixir
Repo.get_by(User, id: id, looking_for_job: true)
```

gemantzu: Thanks :)

PR #599, opened 2020-03-09
URL should be capitalized as it's an acronym

jordelver:
This PR:

* Capitalizes the word URL since it's an acronym
* Removes some unused translations

I've not used `gettext` before so I think this is correct.

I've also removed some translations that aren't required any more since the [job show template was removed in a previous PR](https://github.com/beam-community/elixir-companies/pull/598/files#diff-af9e61a4dc933beb71cff26f5e10fe17).

| Before  | After |
| ------------- | ------------- |
| before  | after  |

### DIFF
--- a/lib/companies_web/templates/job/form.html.eex
+++ b/lib/companies_web/templates/job/form.html.eex
@@ -20,7 +20,7 @@
   </div>
 
   <div class="field">
-    <%= label f, :url, gettext("Url"), class: "field" %>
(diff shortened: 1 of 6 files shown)


doomspork:
@jordelver thanks again for your changes!

jordelver:
@doomspork thanks for merging :)

doomspork:
@jordelver thanks for the contributions both here and at @elixirschool!

gemantzu:
Thanks @jordelver !

PR #605, opened 2020-04-29
updates phoenix to 1.5

gemantzu:


doomspork:
@gemantzu now you can look at enabling the LiveDashboard. If it doesn't have destructive actions (ie it's read-only) maybe we can make it accessible to folks logged in.

PR #607, opened 2020-04-30
Live dashboard

gemantzu:


lib/companies_web/router.ex
```diff
@@ -34,6 +35,12 @@ defmodule CompaniesWeb.Router do
     end
   end
 
+  scope "/dashboard", CompaniesWeb do
+    pipe_through [:browser, :auth]
+
+    live_dashboard "/"
```
SophieDeBenedetto, line 41: We're not specifying a metrics module nor is there one defined in this codebase so far. Should we open a separate issue to set that up? Without it, I don't know that our dashboard will be all that illuminating. I'd be happy to take on that piece of work btw 😄

doomspork: I noticed that when I logged in. Any feedback you have would be great @SophieDeBenedetto! 

If the performance of the dashboard is terrible, I will evaluate options for moving off Heroku,

SophieDeBenedetto: Yep we can enable the `metrics` and `request logger` dashboard tabs with a little work outlined in the docs. I put in an issue here and assigned myself https://github.com/beam-community/elixir-companies/issues/608

I should be able to get to it this week or early next!

gemantzu: @SophieDeBenedetto I don't really know that much about the dashboard yet, so didn't enable anything fancy. You are welcome to tackle all that and enlighten us with your knowledge :)

PR #612, opened 2020-05-14
Add pagination to company overview

maartenvanvliet:
Adds pagination to company search/overview

Could still use some work, maybe someone else can pick that up?
  * Adding/removing event listeners for the edit/delete menu's on the card (but this can also be moved to the company#show view. Slightly easier)
  * Handling of signed_in status in socket
  * Move form fields to changeset/form helpers
  
If all views are converted to LiveView, the scrivener dependency can also be dropped.


lib/companies_web/live/company.ex
```diff
@@ -0,0 +1,79 @@
...
+
+    socket
+    |> assign(
+      companies: companies,
+      page: 1,
+      total_pages: companies.total_pages,
+      text: params["search"]["text"],
+      industry_id: params["search"]["industry_id"]
+    )
+    |> assign(update: "replace")
```
gemantzu, line 73:
```elixir
assign(socket,
  companies: companies,
  page: 1,
  total_pages: companies.total_pages,
  text: params["search"]["text"],
  industry_id: params["search"]["industry_id"],
  update: "replace"
)
```
Make it a signle assign call maybe?

maartenvanvliet: Sure, covered it in this commit https://github.com/beam-community/elixir-companies/commit/9e8407437513794a3de0cc6daa03e5221e498b47 but it's not showing up in the PR

maartenvanvliet:
Yes that would be cool. Should be fairly similar procedure as the company list view.

doomspork:
@gemantzu could you give this another look now that it's been updated?

gemantzu:
@doomspork the only change seems to be a bump in a js package, nothing else. still seems fine.

gemantzu:
@maartenvanvliet can you update your branch once more so I can recheck this weekend and merge? thanks :)

lib/companies_web/helpers/user_helpers.ex
```diff
@@ -8,7 +8,11 @@ defmodule CompaniesWeb.UserHelpers do
     end
   end
 
-  def current_user(conn), do: conn.assigns[:current_user]
+  def current_user(%{assigns: %Phoenix.LiveView.Socket.AssignsNotInSocket{}}), do: false
```
tajchumber, line 11: This should return `nil`, otherwise, [signed_in/1](https://github.com/beam-community/elixir-companies/blob/538c03d081d3cdc3981689e7b5710ff384acb739/lib/companies_web/helpers/user_helpers.ex#L17) function will return true even if the user is not logged in

maartenvanvliet:
Would be great if you can take it over

maartenvanvliet:
It fails CI on master, I think it lost a commit on https://github.com/bglusman/phoenix_live_dashboard

PR #614, opened 2020-06-03
Upgrade elixir

bglusman:
Heroku build failed because some features in master were from newer Elixir version, 1.8, so just bumped to latest 1.10


bglusman:
@maartenvanvliet yeah, it's actually already live on Heroku and this just gets confirmed working fix into master...  though there's something off with git history on Heroku, so we'll see what happens when we merge this, version on Heroku is a rebase off of Heroku master, so not the same SHA as this commit.

PR #615, opened 2020-06-03
Update and improve live dashboard

bglusman:
With this change we'll have history on all the metrics except Phoenix tab... could add those if we want, but I'd already done the VMHIstory module in my LiveDashboard fork, so made sense to bring here....


bglusman:
Failure is from a warning on circular_buffer that was hidden until we hit Elixir 1.10, trouble is though there's a fix in master its not deployed yet via Hex...  could point to git master to fix, or go back to Elixir 1.9 I suppose, both easy fixes, I assume we prefer the former but could also swap out to a different circular buffer implementation I suppose, there are several available.

PR #616, opened 2020-06-05
Force ssl redirect

BobbyMcWho:
The site does not currently force ssl redirects when accessed via http.

It should be noted that `COMPANIES_SSL_KEY_PATH` and `COMPANIES_SSL_CERT_PATH` would need to be set at build time in order for The get_env calls to properly set the keyfile and certfile locations. I believe you have this deployed to heroku, but I don't know enough about heroku deployments to say exactly what these paths should be.


doomspork:
Oh boy. I am embarrassed I missed the configuration for this :cold_sweat: @BobbyMcWho I updated this in Cloudflare using a Page Rule.

image

I appreciate you bringing this to my attention and I'm sorry we won't need this PR. 

I noticed you're a maintainer of omniauth, we're always looking for folks to help with ueberauth if you're interested :grin:

PR #617, opened 2020-06-10
feat: Add historical data to Phx metrics

bglusman:
Also DRY up setup_handlers/children and add two new charts to Phx metrics.

### DIFF
--- a/lib/companies_web/phoenix_history.ex
+++ b/lib/companies_web/phoenix_history.ex
@@ -1,4 +1,7 @@
 defmodule CompaniesWeb.PhoenixHistory do
+  @moduledoc """
+  GenServer to handle historical data for Phoenix telemetry
+  """
   use GenServer
 
   @endpoint_event [:phoenix, :endpoint, :stop]


gemantzu:
Ahm, we do ask for more from GitHub, it's because of the profile page. 
[user.ex](https://github.com/beam-community/elixir-companies/blob/master/lib/companies/schema/user.ex)
[Auth Controller](https://github.com/beam-community/elixir-companies/blob/2f631177765d6daae999f8e23673b23dd21603cf/lib/companies_web/controllers/auth_controller.ex#L9)
Do you think there is a way to ask it dynamically, IE only if someone wants to be a Developer for hire?

doomspork:
@gemantzu we only ask for a single permission. https://github.com/beam-community/elixir-companies/blob/2f631177765d6daae999f8e23673b23dd21603cf/config/config.exs#L42

gemantzu:
I am sorry, I am noob on these permissions. Should we change them because we get more data from Github?

PR #618, opened 2020-06-11
feat: Convert to new historical_data API

bglusman:
After some feedback and ideas from José yesterday I flipped around where and when some of the processing happens for the metrics which should both decrease memory consumption a lot for some metrics, and also leant itself to a nicer API where we can declare the metrics we want history for by just passing them as a list to this new more general/content agnostic GenServer module.  I think overall it's a pretty big win, and will lend itself better to making a new package that relies on dashboard (or for now on my fork of it) and people can get history without having to copy/paste the code templates.  🎉  So I may start on that package/could wait on merging this if you like until I'm done that as other than memory improvement this doesn't change behavior at all, but thought I'd throw the code up for review while I decide next step (this is almost identical now to what's in my live_dashboard fork in dev.exs)


doomspork:
@bglusman can you look into the test failures? The changes otherwise look good to me. So excited to see this project being utilized for the greater good :tada:

bglusman:
Thanks Sean!  Was just a warning for not grouping clauses of a fn together.

bglusman:
Since this wasn't merged yet, and I was working on a proper hex package others can use easily, I switched this PR over to use the new package, though since hex can't use git dependencies, can't publish it until fork is merged into LiveDashboard upstream... but should work the same as previous commit, just, less code in this repo! 🎉

PR #619, opened 2020-06-27
Fix build

tajchumber:
Build fixed by upgrading live_dashboard_history dependency


bglusman:
Oh!  Thank you, I didn’t realize I broke the build with my changes!  I thought I’d kept a branch Thay would be stable as the dependency bit I think I forgot that when merging in stuff the other day!

bglusman:
We can actually remove circular_buffer from lockfile also if you want, I moved to cbuf so it was a) a behavior and b) could easily accommodate Redis or other storage backing mechanisms, but I’ll merge this in either way in a bit with or without that change

tajchumber:
@bglusman I’ve removed unused deps from the lock file now! Can you have a look again?

PR #621, opened 2020-07-14
chore: Update LiveDashboardHistory to hex package

bglusman:


doomspork:
Thank you @bglusman

bglusman:
My pleasure!

bglusman:
Hmm..  @doomspork I'm not sure that's working right, it deployed and it shows 0.2.7 on live dashboard, but I can't click through into metrics...  may need to roll back, I dunno, or see if it works for you locally?  I'm having trouble on both master before and after that merge, so maybe something funny with my local config

bglusman:
I rolled back the deploy on Heroku, when I run master locally it looks like a LiveView issue... this branch did upgrade LiveView to 0.14.1 so I think there must be some issue there... not sure, will update when I figure out.  Though even the rollback doesn't seem to be doing history properly, so, I don't know what's going on

PR #625, opened 2020-07-28
Admin panel

gemantzu:
Needs refinement. I am considering adding LiveView on the list as well for easier access.


doomspork:
@bglusman I wanted to add @nathanl an admin and realized that requires I manually modify the DB. I mentioned to @gemantzu that we should make it easier to add folks to help approve etc.

Where I asked if @nathanl was interested: 
https://elixirforum.com/t/please-add-your-company-to-elixir-companies-com/33256/1

doomspork:
@gemantzu what do you think about sending the user an email to notify them of their admin status? Too much?

gemantzu:
I like it!

bglusman:
Sorry forgot to respond here.... seems reasonable but also I know there’s work in live_dashboard to support admin stuff and custom panels... but I guess since ours is open to the world that wouldn’t be as good a fit, unless we added a second dashboard not open to the world for admins only.  Also not sure whether that can cover our needs yet but maybe worth exploring/I’d be happy to pair or look at with someone if anyone else is interested in trying that route.

doomspork:
@bglusman you know I :100: support you and all your LiveDashboard efforts. If we need to make an admin protected dashboard to feel out some features, let's do it!

bglusman:
Actually, I was getting ahead of myself there, so far they're just doing refactors in support of general admin pages, the feature itself isn't even started yet as best I can tell, I'll keep in mind whenever it is available though!

gemantzu:
@doomspork do you think we should merge this then and come back later if possible?

doomspork (approved):

PR #627, opened 2020-08-04
Search by location together with the text search

cristineguadelupe:
I implemented the search by location together with the search by text, but I do not think that is the best solution. I think about creating a new "by location" search field and renaming the "text" field to "By name". Tell me what you think!

### DIFF
--- a/lib/companies/companies.ex
+++ b/lib/companies/companies.ex
@@ -57,7 +57,7 @@ defmodule Companies.Companies do
   defp query_predicates({"text", text}, query) do
     text = String.trim(text)
 
-    from c in query, where: ilike(c.name, ^"%#{text}%")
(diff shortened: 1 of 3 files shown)


bglusman:
Yeah I think this is fine, maybe worth adding some hover text or other guidance that the search will search either field but don’t think need to add a separate input

gemantzu:
Maybe just add some placeholder text?

cristineguadelupe:
I've added the placeholder text and also increased the text input size a little bit only for desktops to fit the entire placeholder.

cristineguadelupe:
I added some tests following the same pattern as the previous tests. I'm new to the Elixir ecosystem, so please let me know if I'm doing something not too idiomatic.

PR #633, opened 2021-04-29
Remove bio field from user

maco:
It's just a duplicate of the description field. If you look at any profile, it gives the same line twice. If you look at the Ueberauth GitHub library, they were already assigning GitHub's bio into description.


doomspork:
Thanks @maco! Looking at this with fresh eyes I think we can likely drop `description` too. I'll have to double check if there's any other fields we could exclude.

@gemantzu any thoughts here? We've talked about overhauling this project a few times so I'm not sure if you've thoughts about this at all.

PR #665, opened 2023-03-02
Backfill companies

samuelpordeus:
Backfill companies from the previous list using a script.

Companies without a proper location are empty right now:
```ex
%{
  city: "",
  state: "",
  country: ""
},
```
wonder if I can use just `nil` instead and also how it would work for remote/distribute companies 🤔


doomspork (approved):
❤️

Issue #668, opened 2023-03-02
Cannot trigger `load_more` event when scroll down


vvuunnqq:
Hi,
Thanks for making this site live again.

I got an issue when I visited [page](https://elixir-companies.com/en/companies) then I scrolled down to get more companies information, but it doesn't trigger any `load_more` event.
I tried to run this repo on my local machine. Usually, I will get `app.css` and `app.js` under `./priv/static/assets` when run `mix phx.server`. But I saw there is no `app.js` in this folder.
So I think maybe `esbuild` has problem.
I'm looking at this but i'm not sure I can fix or not.

vvuunnqq:
hi @doomspork,
could you please review my PR?
This is issue is resolved with my commit on local machine. I'm not sure If I miss somethings for product configuration, due to I don't have experience to deploy production.
Thanks

PR #670, opened 2023-03-03
Fix issue #668: cannot load_more companies

vvuunnqq:
- Fix: generate + load app.js
- Fix: page number type doesn't match with `pagination_params` func


doomspork (approved):
Thank you and welcome @vvuunnqq!! I just pulled this down and verified it works as expected 🎉 

Look forward to your help making elixir-companies.com the best resource it can be.

PR #684, opened 2023-04-02
Fix issues with sorting

doomspork:
Noticed that the change deployed earlier did nothing. Started messaging with tests and confirmed they weren't actually testing the functionality as expected.

Updated the test data so the test would fail. Updated the sorting functionality to ensure it passed.

### DIFF
--- a/lib/companies/helper.ex
+++ b/lib/companies/helper.ex
@@ -10,8 +10,9 @@ defmodule Companies.Helpers do
   end
 
   def sorted_list(list, params) do
-    {field, direction} = sorting_params(params)
-    Enum.sort_by(list, &Map.get(&1, field), direction)
+    params
+    |> sorting_params()
+    |> sort_by(list)
   end
 
   def paginated_list(list, params) do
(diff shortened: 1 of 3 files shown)


test/companies/companies_test.exs
```diff
@@ -13,69 +13,10 @@ defmodule Companies.CompaniesTest do
     end
   end
 
-  # @test "filters companies by industry" do
```
doomspork, line 16: Just a nitpicl but I removed all of this because we have it in source control and I don't love having huge blocks of commented code living for long periods in any codebase.

test/companies/companies_test.exs
```diff
@@ -13,69 +13,10 @@ defmodule Companies.CompaniesTest do
...
-  #      assert %{entries: [%{name: "ALPHA"}, %{name: "AUSTO"}]} = Companies.all(%{"search" => %{"text" => " aust"}})
-  #      assert %{entries: [%{name: "ALPHA"}, %{name: "AUSTO"}]} = Companies.all(%{"search" => %{"text" => " aust "}})
-  #    end
-  #  end
-
   describe "recent/1" do
     test "retrieves a paginated list of companies sorted by inserted datetime (desc)" do
       assert %{entries: entries} = Companies.recent()
-      assert [%{name: "ALPHA"}, %{name: "ZULU"}] = entries
+      assert [%{name: "ZULU"}, %{name: "ALPHA"}] = entries
```
doomspork, line 19: The A-Z sort was _actually_ at play here. Change the fixture dates to be reversed so that we could _really_ ensure this was working as expected.

test/companies/companies_test.exs
```diff
@@ -84,9 +25,9 @@ defmodule Companies.CompaniesTest do
       assert %{name: "ZULU"} = Companies.get!("zulu")
     end
 
-    # test "does not retrieve deleted record" do
-    #  assert_raise Ecto.NoResultsError, fn -> Companies.get!(company.id) end
-    # end
+    test "returns nil if company is missing" do
+      refute Companies.get!("beta")
+    end
```
doomspork, line 30: Removed the commented out test and updated it with a test that tested the current functionality

lib/companies/helper.ex
```diff
@@ -10,8 +10,9 @@ defmodule Companies.Helpers do
   end
 
   def sorted_list(list, params) do
-    {field, direction} = sorting_params(params)
-    Enum.sort_by(list, &Map.get(&1, field), direction)
+    params
+    |> sorting_params()
+    |> sort_by(list)
```
doomspork, line 15: The downside to using a list of values vs DB for the data source is that sorting is not as capable. Works for now but may need to rethink in the future.

yordis: This is fun!

yordis (approved):

PR #685, opened 2023-08-23
Rename Margo Bank with Memo Bank

mgalibert:


doomspork:
Thank you @mgalibert!

PR #687, opened 2023-09-07
Add Dutch translation

Hermanverschooten:
Even though most are not used anymore...


doomspork:
> Even though most are not used anymore...

Thank you @Hermanverschooten! Let me see if I can make time this weekend to take a look at gettext and get things updated to the latest and greatest.

doomspork (approved):

PR #699, opened 2024-09-07
Upgrade: elixir 1.17 otp 27 node 22

begedin:
I tried to review (manually test) the currently open dependabot PR and found myself unable to run the project on an M2 mac.

So I invested about an hour to try and upgrade the thing to the latest and greatest

- elixir to 1.17.2
- otp to 27.0.1
- node to 22.8

Outside of that, i had to manually bump `httpoison` as `httpoison -> hackney -> ssl_verify_fun` refused to compile.
I also did a `mix hex.update --all`

Finally, I bumped everything in the `Dockerfile` to and then tried to deploy on a fly machine of my own. I found the min required version for dart_sass went up to 1.55 on linux, so since i was already bumping that, I went with the latest version.

And then after the PR failed, I realized I also had to upgrade credo cause it wouldn't run

After setting the necessary secrets, the deployment worked and everything seems to be in order. I took my own app down after, but I can put it back up if anyone wants to manually test.


config/prod.exs
```diff
@@ -11,7 +11,7 @@ import Config
 # before starting your production server.
 config :companies, CompaniesWeb.Endpoint,
   http: [:inet6, port: System.get_env("PORT") || 8080],
-  url: [scheme: "https", host: "elixir-companies.com"],
+  url: [scheme: "https", host: "bego-elixir-companies.fly.dev"],
```
doomspork, line 14: We need to change this back.

```suggestion
  url: [scheme: "https", host: "elixir-companies.com"],
```

doomspork (approved):

config/prod.exs
begedin: Ooops, sorry, missed it. Thank you for correcting it.

doomspork: No problem @begedin! Teamwork makes the dream work 🎉 Thanks for helping out. Need to make time in 2025 to give this site a good overhaul. Your thoughts and feedback are welcome!

PR #705, opened 2024-10-21
fix: rm non-functional industries select, make search box bigger

jsegal205:
The industries dropdown was non functional as the select box was empty - [code reference](https://github.com/beam-community/elixir-companies/blob/main/lib/companies/industries.ex#L36-L38), and the search box will search on industry or company name - [code reference](https://github.com/beam-community/elixir-companies/blob/main/lib/companies/helper.ex#L15). 

Changes:
- removes industries select control
- removes class that shrunk the search input to itty bitty
- update search placeholder text to be accurate of what will be searched

## before
Screenshot 2024-10-21 at 3 02 42 PM

## after
Screenshot 2024-10-21 at 3 02 55 PM


doomspork (approved):
Thank you @jsegal205!

PR #708, opened 2024-10-22
Update .tool_versions

nathanl:


doomspork (approved):

jsegal205:
IMO this will / should be managed in the future by [common-config](https://github.com/beam-community/common-config).

PR #671, opened 2023-03-04
Update industries to match GICS

doomspork:
Updating all of the industries to match the 11 outlined by the GISC and currently in the industries.ex file.

https://www.msci.com/our-solutions/indexes/gics

PR #711, opened 2024-10-22
Validate industries GHA

jsegal205:
Adds custom validation so that new / updates to company files add a standard for industries. 

Supports #671 in the long term

### DIFF
--- a/lib/companies/industries.ex
+++ b/lib/companies/industries.ex
@@ -2,8 +2,8 @@ defmodule Companies.Industries do
   @moduledoc """
   The Industries context.
 
-  Derived from the 11 sectors of the Stock Market
-  https://www.nasdaq.com/articles/the-11-sectors-of-the-stock-market-their-biggest-etfs-2021-05-30
+  Pulled from the Global Industry Classification Standard (GICS)
+  https://www.msci.com/our-solutions/indexes/gics
   """
 
   @industries [
@@ -14,9 +14,9 @@ defmodule Companies.Industries do
     "Financials",
     "Health Care",
     "Industrials",
+    "Information Technology",
     "Materials",
     "Real Estate",
-    "Technology",
     "Utilities"
   ]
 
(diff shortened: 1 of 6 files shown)


nathanl (approved):

lib/mix/tasks/validate_company_file.ex
```diff
@@ -0,0 +1,38 @@
...
+  Validates a filepath to a company.
+
+      $ mix validate_company_file priv/companies/abc.exs
+
+  This is to help with data consistency.
+  """
+
+  @impl Mix.Task
+  @doc false
+  def run([file_path]) do
```
btkostner, line 17: It might be worth doing a `Path.wildcard` so the function can run over all the files. It would be simpler to do it in Elixir than doing it via a bash shell, and it would be easier for devs to run locally.

lib/companies/industries.ex
```diff
@@ -2,8 +2,8 @@ defmodule Companies.Industries do
   @moduledoc """
   The Industries context.
 
-  Derived from the 11 sectors of the Stock Market
-  https://www.nasdaq.com/articles/the-11-sectors-of-the-stock-market-their-biggest-etfs-2021-05-30
+  Pulled from the Global Industry Classification Standard (GICS)
+  https://www.msci.com/our-solutions/indexes/gics
```
btkostner, line 6: @doomspork Are you ok with this change? Seems simple enough.

lib/mix/tasks/validate_company_file.ex
jsegal205: I was hesitant to run it over all the files as a good majority of the existing company will likely need to be updated. My thought was adding this so that new or updates to company files could have some data consistency for industries.

This was my attempt at something similar to an ecto changeset validation rule without having or adding ecto back to the project. 

I'm open to other ideas or approaches.

lib/companies/industries.ex
doomspork: Yes, I started on this a long time ago in #671

doomspork (approved):

PR #715, opened 2024-10-29
feat: Enhance validations and mix task

doomspork:
This is a proof-of-concept. @jsegal205 @btkostner thoughts?

image


lib/companies/schema/company.ex
```diff
@@ -1,27 +1,33 @@
...
   @moduledoc false
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
   alias Companies.Industries
 
-  # @enforce_keys [:industry, :last_changed_on, :location, :name, :website]
+  schema "companies" do
+    field(:blog, :string)
```
doomspork, line 10: Need to update the .formatter.exs if we keep this.

lib/companies/schema/company.ex
```diff
@@ -1,27 +1,33 @@
...
-    :location,
-    :name,
-    :website
-  ]
+    timestamps()
+  end
 
   def build(_filename, attrs, _body) do
     struct!(__MODULE__, Map.to_list(attrs))
   end
```
doomspork, line 25: Let's rethink this.

mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule Companies.MixProject do
     [
       app: :companies,
       version: "0.1.0",
-      elixir: "~> 1.17",
+      elixir: "~> 1.16",
```
doomspork, line 8: I'll undo this.

mix.exs
```diff
@@ -77,7 +77,8 @@ defmodule Companies.MixProject do
       {:floki, ">= 0.0.0", only: :test},
       {:phoenix_live_reload, "~> 1.2", only: :dev},
       {:dart_sass, "~> 0.5", runtime: Mix.env() == :dev},
-      {:parallel_stream, "~> 1.1", only: [:dev, :test]}
+      {:parallel_stream, "~> 1.1", only: [:dev, :test]},
+      {:ecto, "~> 3.12", only: [:dev, :test]}
```
doomspork, line 81: 🔤

doomspork:
I also updated Zipbooks to trigger the GHA @jsegal205 made

lib/companies/schema/company.ex
```diff
@@ -1,27 +1,33 @@
 defmodule Companies.Schema.Company do
   @moduledoc false
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
   alias Companies.Industries
 
-  # @enforce_keys [:industry, :last_changed_on, :location, :name, :website]
+  schema "companies" do
```
btkostner, line 9: Odd thought, but what about adding a `file_path` attribute to the struct? This will allow us to add validation messages on it, as well as allowing us to do fun future things like adding a "Quick Edit" button in the web UI that points to GitHub edit page.

lib/mix/tasks/validate_companies.ex
```diff
@@ -0,0 +1,74 @@
...
+        :error
+
+      {:error, reason} ->
+        IO.puts("❌ #{file_path} failed: #{reason}")
+        :error
+
+      _ ->
+        :error
+    end
+  end
```
btkostner, line 58: If we do the above with `file_path`, I'd simplify this function and have it always return a changeset. If we can't parse the file, then we create a new empty changeset with an error on `file_path`.

This would allow us to separate the validation logic and the display validation logic.

lib/companies/schema/company.ex
btkostner: Yeah, seems weird to have a build function with two of the params not used.

lib/companies/schema/company.ex
```diff
@@ -1,27 +1,33 @@
...
+    field :blog, :string
+    field :description, :string
+    field :github, :string
```
btkostner, line 12: We should validate the format to a URL with `github.com` as the host, and no `/` in the path.

lib/companies/schema/company.ex
```diff
@@ -1,27 +1,33 @@
...
+    field :industry, :string
+    field :jobs, {:array, :string}
+    field :last_changed_on, :date
+    field :location, :map
```
btkostner, line 16: Also possibly a nested struct with it's own validation.

btkostner (commented):
I like. I added some comments about maybe adding more info. Take it or leave it (my brain might be fried)

test/schema/company_test.exs
```diff
@@ -1,12 +1,26 @@
...
-      refute Company.validate_industries(%Company{industry: "Non-existant"})
+  @attrs %{
+    name: "Acme Corp",
+    website: "https://example.com/",
+    github: "",
+    industry: "Information Technology",
+    location: %{"city" => "City", "state" => "State", "country" => "Country"},
+    description: "Description of Acme Corp goes here.",
+    last_changed_on: "2021-01-01"
+  }
```
jsegal205, line 14: Would it make sense to use either the [alpha.exs](https://github.com/beam-community/elixir-companies/blob/main/test/support/companies/alpha.exs) or [zulu.exs](https://github.com/beam-community/elixir-companies/blob/main/test/support/companies/zulu.exs) factories (with a bit of updating to make them valid)?

lib/companies/schema/company.ex
```diff
@@ -1,27 +1,33 @@
 defmodule Companies.Schema.Company do
   @moduledoc false
+  use Ecto.Schema
+
+  import Ecto.Changeset
```
jsegal205, line 5: ❤️ 

I briefly thought about adding ecto and using a changeset to do the validations in #711, but I wasn't sure about the overhead of adding ecto when it wasn't used in tandem with a DB. Glad you went the way of ecto.

lib/companies/schema/company.ex
jsegal205: For future, we could make this more generic of `field :open_source_link, :string` to be inclusive of gitlab or bitbucket.

test/schema/company_test.exs
doomspork: I thought about it and I want to clean all of that up. Alpha and Zulu exist to be valid and loaded by the nimble publisher for some other tests, they wouldn't work for the invalid companies so need to solve for that.

lib/companies/schema/company.ex
doomspork: We don't need to add a DB adapter to Ecto and we can specify it's dev/test only to eliminate the production overhead.

lib/companies/schema/company.ex
doomspork: Agreed. Location has definitely been a toughy.

lib/mix/tasks/validate_companies.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Mix.Tasks.ValidateCompanies do
```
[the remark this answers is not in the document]
doomspork: Yes but the tasks + action need a little more work, especially when there are non-company files modified.

test/schema/company_test.exs
doomspork: The other reason I didn't use one here is I wanted the parameters to be clear in the test not hidden in a factory or fixtures that's loaded into a variable and passed in.

jsegal205: Yeah that makes sense to me. My brain leans towards DRY things but explicit and obvious makes sense. Let me know if I can help with any of the cleanup.

lib/mix/tasks/validate_companies.ex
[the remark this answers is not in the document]
jsegal205: FYI I made a bunch of updates to the current GHA file for validating companies with https://github.com/beam-community/elixir-companies/pull/717 . This _SHOULD_ only pick company related files in pull requests.

PR #719, opened 2024-10-31
Update naming of company industry gha

jsegal205:


.github/workflows/validate_company_industry.yml
```diff
@@ -2,11 +2,10 @@ name: Validate Company Industry
 
 on:
   pull_request:
-    paths:
-      - "priv/companies/*.exs" # only run on changes to company files
```
jsegal205, line 6: As the workflow is required, I removed the path filtering for when the workflow is ran otherwise the required check would remain in pending state [per the docs](https://docs.github.com/en/enterprise-cloud@latest/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks).

This path protection is also listed in the changed file job - https://github.com/beam-community/elixir-companies/blob/main/.github/workflows/validate_company_industry.yml#L39

btkostner (approved):

PR #720, opened 2024-11-01
support GHA on push to main and adding to merge queue

jsegal205:
ISSUE: 
Screenshot 2024-11-01 at 11 08 30 AM

Adding things to the merge queue didn't run the validate company GHA. 

Github docs on [`merge_group`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#merge_group).

### DIFF
--- a/.github/workflows/validate_company_industry.yml
+++ b/.github/workflows/validate_company_industry.yml
@@ -2,6 +2,10 @@ name: Validate Company Industry
 
 on:
   pull_request:
+  push:
+    branches:
+      - main
+  merge_group:
 
 jobs:
   validate:


.github/workflows/validate_company_industry.yml
```diff
@@ -2,6 +2,10 @@ name: Validate Company Industry
 
 on:
   pull_request:
+  push:
+    branches:
+      - main
```
jsegal205, line 7: I'm not sure if this is FULLY needed but I added it for good measure just in case.

doomspork (approved):

PR #721, opened 2024-11-01
[feat]: add Beam community common config

jsegal205:
Implement https://github.com/beam-community/common-config


.github/workflows/common-config.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .tool-versions
+
+      - name: Setup Elixir
+        uses: stordco/actions-elixir/setup@v1
+        with:
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
```
jsegal205, line 40: This doesn't specify the elixir and erlang versions here. [By default this action will read from the .tool-versions file](https://github.com/stordco/actions-elixir/blob/main/setup/README.md#usage) and this repo uses differ node, elixir and erlang version than what is noted in common-config repo file.

@btkostner is there some documentation we can / should add to the common-config repo for how to override this file so it doesn't get lost with the periodic syncs? Or is the comment at the top of this file a bit of a scare tactic?

btkostner: This is kinda intentional because we might need to load dependencies and changing versions can break that. With that said, it's not a great solution because it also means we don't know what elixir version our common-config scripts will run against.

doomspork (approved):

PR #723, opened 2024-11-08
rename caredox to schoolcare bc the website and linkedin both have that redirect

maco:
The website that was listed and the linkedin page both have this redirect, so updating the url and name to reflect 2024

### DIFF
--- a/priv/companies/caredox.exs
+++ b/priv/companies/caredox.exs
@@ -1,5 +1,5 @@
 %{
-  industry: "Healthcare",
+  industry: "Health Care",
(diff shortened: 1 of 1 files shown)


priv/companies/caredox.exs
```diff
@@ -1,14 +1,14 @@
 %{
   industry: "Healthcare",
```
jsegal205, line 2:
```suggestion
  industry: "Health Care",
```

jsegal205 (approved):
TY for the update @maco

PR #724, opened 2024-11-08
rename Divvy to Bill

maco:
Bill.com bought Divvy


priv/companies/divvy.exs
```diff
@@ -1,14 +1,14 @@
 %{
   industry: "Financial Services",
```
jsegal205, line 2:
```suggestion
  industry: "Financials",
```

jsegal205 (approved):
TY for the update @maco

PR #725, opened 2024-11-08
remove letote

maco:
they went bankrupt, got bought, then the buyer was shut down 

https://en.wikipedia.org/wiki/Le_Tote


btkostner (approved):
Thank you for the update!

PR #728, opened 2024-11-15
chore: common config credo fixes

jsegal205:
Fixes credo issues with common config that was initially pulled over in https://github.com/beam-community/elixir-companies/pull/726


btkostner (approved):
Nice!

PR #732, opened 2024-12-05
Add All Aboard

PatrikStenmark:
## New Company Prerequisites

- [x] Industry must be one of the 11 sectors of the [GICS](https://www.msci.com/our-solutions/indexes/gics).
- [x] The description makes some mention of Elixir usage.

PR #733, opened 2024-12-06
Default industry

PatrikStenmark:
In #732 I just assumed that the `industry` I got from the mix task was a valid alternative, which it wasn't. I'm thinking that it makes sense to have a valid alternative.

I also add a reference to the `validate_company_file` so one can run the validation before creating a PR.


doomspork (approved):
Great catch @PatrikStenmark 🙏

PR #757, opened 2025-04-08
feat: add Keila

wmnnd:
## New Company Prerequisites

- [X] Industry must be one of the 11 sectors of the [GICS](https://www.msci.com/our-solutions/indexes/gics).
- [X] The description makes some mention of Elixir usage.

### DIFF
--- a/priv/companies/keila.exs
+++ b/priv/companies/keila.exs
@@ -9,5 +9,6 @@
   last_changed_on: ~D[2025-04-08],
   description: """
   Keila is an Open Source email newsletter tool and an alternative to MailChimp. It’s available as a managed service (Keila Cloud) but can also be self-hosted.
+  Keila is built using the Phoenix framework and uses various common Elixir libraries, including Ecto, Oban, and Earmark.
   """
 }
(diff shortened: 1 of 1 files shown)


priv/companies/keila.exs
```diff
@@ -0,0 +1,14 @@
+%{
+  industry: "Marketing/Sales",
```
jsegal205, line 2: @wmnnd can you update this to align the industry with the set of industries that we support? You can find a list of supported industries in https://github.com/beam-community/elixir-companies/blob/main/lib/companies/industries.ex#L9-L23

wmnnd:
Sure thing, I just changed it!

By the way, the reason why I choose `Marketing/Sales` is because there are a couple of other companies using this tag as their `"industry"`: https://github.com/search?q=repo%3Abeam-community%2Felixir-companies%20Marketing%2FSales&type=code

doomspork:
@wmnnd we are working to update older listings to use the smaller set of industries, this got _very_ out of hand before with many invented industries that then degraded the search ability on the site.

doomspork (approved):

